PR #21, opened 2016-10-07
Add `IsEqual` to function pool

yunchih:


ihhub:
Could you please make some changes to follow overall project structure:
1. Add a test for second variation of function
2. Put functions in alphabet order
3. Use spaces instead of tabs to format comments
4. For unit test I would recommend to generate 2 different images with known pixel intensities and compare values then check return value from isEqual() function. Now the test checks the same image

yunchih:
1. `Makefile.inc` was intended to reduce rules repetitiveness, but since users find it easier to have each example as self-contained as possible, that's do as what you said.
2. 3. 4. 5. thanks for the reminder.  These do seem like stupid mistakes.


By the way, I can link with `clang++` fine, but cannot do so with `g++`, which complains, all within the file `function_pool.cpp`:

```
undefined reference to `void Image_Function::ParameterValidation<(unsigned char)1>(Bitmap_Image::BitmapImage<(unsigned char)1> const&, unsigned int, unsigned int, unsigned int, unsigned int)'
```
The reason is that I compile each `.cpp` into object files and link them at the last stage.
The linking only works if I move these template functions from `image_function.cpp` to `image_function.h`, so that when `function_pool.cpp` is being compiled, function  `ParameterValidation` is in scope.

ihhub:
1. unit_test_function_pool.cpp file error of compilation, line 425: generateRoi( input, roiX, roiY, roiWidth, roiHeight );
2. Adding headers to project would help to solve compilation problem with template functions
The most important could you please make pull request as 1 commit. It would be difficult to trace problems if we have so many commits just to fix 1 issue.

yunchih:
Sorry about messing things up.
I've pruned the commit associating with Makefile in exampls and squashed other commits into one.

ihhub:
It's okay. No problem. I will review the code soon.

yunchih:
I have no problem compiling with clang++ 3.6 on Ubuntu 15.10.
Here is the running log: http://codepad.org/dlZr5fv6
Could you provide the failing compile log?

ihhub:
From your log you do **NOT** run Function Pool tests. Include these tests into unit_tests.cpp, compile and run it:
````cpp
// This application is designed to run unit tests on penguinV library
#include <iostream>
#include "unit_test_bitmap.h"
#include "unit_test_blob_detection.h"
#include "unit_test_framework.h"
#include "unit_test_function_pool.h"
#include "unit_test_image_buffer.h"
#include "unit_test_image_function.h"

int main()
{
	// The main purpose of this application is to test everything within library
	// To do this we need an engine (framework) and a bunch of tests

	// We create a framework
	Unit_Test::UnitTestFramework framework;

	// We add tests
	Unit_Test::addTests_Bitmap        ( framework );
	Unit_Test::addTests_Blob_Detection( framework );
	Unit_Test::addTests_Image_Buffer  ( framework );
	Unit_Test::addTests_Image_Function( framework );
	Unit_Test::addTests_Function_Pool( framework );

	// Just run the framework what will handle all tests
	return framework.run();
}
````

yunchih:
Thanks for the head up~  I have pushed the fix.
However, the two newly-added tests still do not succeed.
There seem to be some unpredictable failure that takes some more debugging.

PR #22, opened 2016-10-11
Makefiles for examples

syclik:
@ihhub:

- I added makefiles for all the examples, except `qt_framework` (I couldn't get that to build)
- in order to get `blob_detection` to build, I had to include `cmath`. See 0ccc908 (it wouldn't build using the instructions in the readme)
- I've relied on `make` as much as possible. Everything is laid out the same and will be easy for you to maintain. It'll also be clear to someone that knows make. It's always `target : dependencies`.
- Each of the executables are named after the C++ file instead of `application`. That allows the makefile to be minimal since make understands that. If you want everything to be copied, it can be done.
- The way it's set up, it will respect the dependencies listed. It doesn't create a dependency tree by running the compiler on the files. This is easier to deal with especially on a smaller project, but you won't have automatic rebuilding when you change a file that's included by one of those files listed.
- This isn't cross platform, but you already have MSVC projects, so I don't think that's a big deal.
- I updated the doc to include `make`.
- There's also a `make clean` target that removes the executable.
- If you want to change the compiler, you can do so when calling make: `make CXX=clang++`

I think that's it for now.

### DIFF
--- a/PenguinV/Library/blob_detection.cpp
+++ b/PenguinV/Library/blob_detection.cpp
@@ -1,3 +1,4 @@
+#include <cmath>
 #include <numeric>
 #include <queue>
 #include "blob_detection.h"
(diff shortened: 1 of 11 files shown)


ihhub:
Hi!

Thanks for your work.    
- `qt_framework` example is made for QT what is different framework. You cannot compile it using `make`    
- C++ code is cross-platform, project files are not. So it is okay

PR #30, opened 2016-10-26
Using vector::empty() instead of vector::clear()

rab-h-sonmai:
I noticed that function_pool::getSum() is using sum.empty()  before returning.
The return type of which is ignored and the comment next to it seem to imply that clear() was the original intention...?


ihhub:
Yes, you are right. Thanks for the fix!

Issue #56, opened 2017-10-01
Static library compilation


ihhub:
As a library the project should be built as a static library as well. We must add necessary project/make files for this.

From design point of view it's a good practice to put project/make files out of main ```PenguinV\Library``` directory. Recommended path for this is ```PenguinV\build\static_library```

platipo:
Correct me if I'm wrong, you want the sources in Library folder to compile into a static library?

ihhub:
Hi, @platipo ,

Yes, you are correct. Ideally all files and folders within ```Library``` folder (including CUDA code as well).

platipo:
Got a problem:
```
g++ -Wall -g -O0 -pthread -std=c++11 -c penguinv/penguinv.cpp -o penguinv/penguinv.o
In file included from penguinv/penguinv.cpp:1:0:
penguinv/cpu_identification.h:42:3: error: #error "Unknown platform"
  #error "Unknown platform"
   ^~~~~
make: *** [Makefile:21: penguinv/penguinv.o] Error 1
```

platipo:
And also:
```
image_function_neon.cpp:1:10: fatal error: arm_neon.h: No such file or directory
 #include <arm_neon.h>
          ^~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:21: image_function_neon.o] Error 1
```
Even though `$ g++ -march=native -dM -E - </dev/null | grep -i arm` returns nothing.

ihhub:
Library has not been tested for a while under Linux so your suggestion could be correct.

Neon is SIMD instruction set which is only for ARM processors. I assume you compile on x86 processor. But I will make a proper changes for this for avoid such problem.

ihhub:
Closing this issue as we have similar issue.

PR #57, opened 2017-10-01
Add Makefile for static library compilation

platipo:
Unfortunately I cannot test it neither with CUDA nor under Windows.

#56

### DIFF
--- a/PenguinV/unit_tests/cuda/unit_test_helper_cuda.cu
+++ b/PenguinV/unit_tests/cuda/unit_test_helper_cuda.cu
@@ -1,5 +1,6 @@
 #include <cuda_runtime.h>
 #include "../../Library/image_buffer.h"
+#include "../../Library/image_exception.h"
 #include "../../Library/image_function.h"
 #include "../../Library/cuda/cuda_types.cuh"
 #include "../../Library/cuda/cuda_helper.cuh"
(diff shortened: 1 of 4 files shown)


ihhub:
Hi, @platipo ,

Thank you for your efforts!

I had a look at your pull request. I have multiple suggestions/remarks:
1. For me compilation failed with such error:
```
mvcc -std=c++11 -c cuda/cuda_device.cu -o cuda/cuda_device.o
make: mvcc: No such file or directory
make: *** [cuda/cuda_device.o] Error 1
```

I tested on MacOS (Unix system) which does not have CUDA library.

2. CUDA compiler is called ```nvcc```, not ```mvcc``` but it doesn't solve the problem above.

3. We should put a makefile out of ```Library``` folder to do not contaminate source folder. I recommend to put makefile under ```build\static_library``` as for now and put all static files there. Ideally copy header files there too because they are needed to use the library.

4. We should include all files inside Library folder without setting specific names of files. Otherwise every time we have to modify makefile once we add/remove some files. This needs some preparations. You didn't include files like ```image_function_sse.cpp``` or ```image_function_neon.cpp``` because of SIMD restrictions. I will change those files in upcoming days that you can include them easily into makefile.

5. Make a pull request for linux build separately. It's a good practice to do separate commits in separate pull request.

platipo:
1. What flags do you use when compiling with `nvcc`? As In stated I cannot verify the CUDA part
2. Done
3. I think it isn't necessary to reinvent the whell, but we should follow a well known [structure](https://stackoverflow.com/questions/1398445/directory-structure-for-a-c-library#1398594). I suggest you to reorganize the codebase
4. I have to wait until you change the structure for SSE, AVX and SIMD and then I can change the makefile
5. You are right, done

PR #62, opened 2017-10-07
Replace uint32_t with size_t

### LINKED ISSUE #61: uin32_t --> size_t for bigger image support
Most of values are uint32_t while we can use size_t which is same unit32_t in 32 bit applications but uint64_t in 64 bit application which gives to us a support of bigger images support.

**Notes for considerations**: main goal is to change Image classes and respective functions but do not touch code which is 3rd party or related to some standards like Bitmap file management code.

platipo:
Resolves #61

### DIFF
--- a/PenguinV/Library/blob_detection.cpp
+++ b/PenguinV/Library/blob_detection.cpp
@@ -20,32 +20,32 @@ namespace
 
 namespace Blob_Detection
 {
-    const std::vector < uint32_t > & BlobInfo::pointX() const
+    const std::vector < size_t > & BlobInfo::pointX() const
     {
         return _pointX;
     }
 
-    const std::vector < uint32_t > & BlobInfo::pointY() const
+    const std::vector < size_t > & BlobInfo::pointY() const
     {
         return _pointY;
     }
 
-    const std::vector < uint32_t > & BlobInfo::contourX() const
+    const std::vector < size_t > & BlobInfo::contourX() const
     {
         return _contourX;
     }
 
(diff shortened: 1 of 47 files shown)


ihhub:
Hi, @platipo . Thanks for efforts. Looks like you just replace `uint32_t` with `size_t` in all files. There are multiple problems:
1. The pull request has conflicts with `master` branch
2. Right indentation in files is gone
3. Bitmap format does not support size_t aka uint64_t
4. Some CUDA operations do not support uint64_t
5. Png format has it's own formats

platipo:
1. We are here to resolve them
2. This isn't a major problem I'll fix it before merging
3. Why is the description not mentioning it?
4. Same as 3
5. Same as 3

And how can I distinguish between `uint32_t` to replace and the ones to keep?

ihhub:
A good written source code contains less mistakes and it is easy to read by others so indentation is important to maintain a good quality code. Obviously if it was just to replace `uint32_t` by `size_t` in all files I wouldn't have created an issue for this because it is just 1 minute job. I will change description of the issue after deeper look what we need and what we do not need to change.

platipo:
I'm not saying that identation isn't important but that I'll fix the identation before merging because it seems that this PR commits will change again and again.
Regarding identation and contributions I saw no `CONTRIBUTING.md` in this repo, I think we can agree that it's necessary so everyone follow the same workflow and coding style. Here is a [fanstatic example](https://github.com/atom/atom/blob/master/CONTRIBUTING.md).

ihhub:
As described in related issue there no necessary need to do this.

PR #112, opened 2018-10-07
Optimized performance of GetGaussianKernel() by taking invariant comp…

### LINKED ISSUE #96: GetGaussianKernel() optimisation
There are multiple calculations inside loop which we could optimise by calculating temporary results.

NFM-8:
…utations out of loops. Closes #96.


ihhub:
Thanks @NFM-8 for contributing to the project!

Issue #132, opened 2018-10-11
Compilation failure due to missing cpuid.h in Linux systems


0x72D0:
device: raspberry 3B
cpu: cortex-a53
linux kernel: 4.9.35-v7+

I have some problem when I try to compile the unit_tests on arm. I get the following error:

![2018-10-11-015434_1824x984_scrot](https://user-images.githubusercontent.com/16087328/46776128-3ccefb00-ccd8-11e8-8e4d-29bfa7e25476.png)

and when I check in the gcc project config folder, arm don't have cpuid.h:

[x86 files](https://github.com/gcc-mirror/gcc/tree/master/gcc/config/i386)
[arm files](https://github.com/gcc-mirror/gcc/tree/master/gcc/config/arm)

Also the -march=native don't work on the raspberry pi, I have to change it to: 
-mcpu=cortex-a53 -mfpu=neon-vfpv4

ihhub:
@0x72D0 please try this:
In `cpu_id_unix.h` file do following changes:
1. replace
`#include <cpuid.h>`
by
`#ifndef __arm__`
`#include <cpuid.h>`
`#endif`

2. Put a whole body of `isSseSupported` and of `isAvxSupported` functions except last line (`return false;`) into
`#ifndef __arm__`
`        function's body`
`#endif`

If this fix would work please upload a new pull request. No left spaces for `#ifndef` and `#endif` (they should start at the beginning of the line).

ihhub:
@0x72D0 did my suggestion for the fix help?

0x72D0:
sorry I didn't find the time to test it yet... I'm gonna test it this weekend

0x72D0:
ok I succeed to compile it with your suggestion, but it dosen't work with the -march=native. It give me the following error (after the make):

![2018-10-12-210645_1824x984_scrot](https://user-images.githubusercontent.com/16087328/46894826-5487ba00-ce65-11e8-8f3c-3cd20fc97c97.png)

but with -mcpu=cortex-a53 and -mfpu=neon-vfpv4 everything work fine.

ihhub:
Let's open another issue for such case as you fixed original. Please have a look into #138.

PR #136, opened 2018-10-12
fix the cpuid problem with arm processor

0x72D0:
PR for #132


ihhub:
Thanks for your efforts to contributing to the project!

Issue #106, opened 2018-10-05
Neon implementation of Sum function


ihhub:
`Sum()` function is not implemented for NEON (SIMD) yet. Please refer to `image_function_simd.cpp` file.

0x72D0:
I would like to learn more about simd. I will try this one!

ihhub:
Just don't forget to add a unit test of your newly created function for verification (one line code in `unit_test_image_function.cpp`).

ihhub:
@0x72D0 you could also have a look at another SIMD related issue. I've added them recently.

0x72D0:
ok thanks for the proposition, but I think I will stay with this issue. I have a raspberry pi 3B at home and the datasheet of the CPU say it has the neon extension set, so I can test the function on my raspberry pi. The only problem is my lack of experience with the arm architecture, so do you have good reference for the neon extension? For some reason every arm reference manual I came across seems outdated...

ihhub:
@0x72D0 you can refer to http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0204j/Bcfjicfj.html (the page could load very slow due to big amount of data).

By the way, please include your new function into performance tests (1 line of code in `performance_test_image_function.cpp`) to see the improvement in speed.

0x72D0:
ok thanks!

PR #145, opened 2018-10-16
add sum to arm neon

0x72D0:
PR for #106 all test unit_pass and performances are good. Here some screenshot ( I disabled a lot of test to avoid useless waiting ):

2018-10-16-023658_1824x984_scrot
2018-10-16-024119_1824x984_scrot

so the normal performance was: 0.16 sec / 0.70 sec / 2.8 sec / 11.12 sec
now the neon performance is:     0.04 sec / 0.17 sec / 1.1 sec / 3.691 sec

### DIFF
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -1039,6 +1039,36 @@ namespace neon
         }
     }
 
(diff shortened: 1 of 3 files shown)


ihhub:
Thanks a lot for the contribution!

Issue #149, opened 2018-10-19
Remove cpu_identification.h inclusion from image_function_simd.h


ihhub:
`#include "penguinv/cpu_identification.h"` is not valid for `image_function_simd.h` and should be moved to `image_function_simd.cpp` file instead. Once it's done we **might** need to modify file inclusions in other files.

Related to #98

PR #153, opened 2018-10-19
Move `cpu_identification` include to from `.h` to `.cpp` file.

dainmiller:
Addresses #149


ihhub:
Thanks a lot for contribution!

Issue #45, opened 2017-07-03
libjpeg support


ihhub:
We have libpng support but not libjpeg. Many images are stored in this format and it would be irrational to do not support it.

Correlates with #7

ngetahun:
Hi @ihhub, I want to get started with this issue. Is it still up for grabs?

ngetahun:
@ihhub I saw that you didn't add any unit test for png support. Should I do the same for jpeg?

ngetahun:
Also, it would be nice to setup some sort of build system (perhaps ```cmake```) or a configuration system (like GNU's ```autoconf```) to check and resolve dependencies.

ihhub:
Ideally we should add unit tests for jpeg. I didn't add unit tests for png because png is not supported on Windows yet (have a look into #41 ).

If you are intending to add jpeg support for Unix OSes as of now then add unit tests under `ifdef` condition for non-Windows OSes. Later we could add Windows code support and remove `ifdef`.

ihhub:
Regarding `cmake`: you are right we should have it too because we cannot build a whole code without checking the presence of certain libraries like `libpng` or `cuda`. Could you please create an issue for this?

ihhub:
I'll create another issue regarding unit tests for `libpng` as well.

ngetahun:
#159

yannlabou:
> Regarding `cmake`: you are right we should have it too because we cannot build a whole code without checking the presence of certain libraries like `libpng` or `cuda`. Could you please create an issue for this?

Yes. #270

PR #159, opened 2018-10-19
Add jpeg support

ngetahun:
Added support for jpeg library


ngetahun:
#45

ngetahun:
@ihhub can you put here what's causing the CI failure.

ihhub:
Hi, @ngetahun .

Did you have a change to have a look into given comments? If you have some misunderstanding or question please do not hesitate to ask. I would like to help :)

ihhub:
I've made some changes in your code. Please have a look into them.

ihhub:
No problem. I just had a free time today so I made certain changes as you might have been busy :)

ngetahun:
@ihhub is there something missing from this PR?

ihhub:
I made changes in your code but I haven't tested them. Additionally, we should address [this comment](https://github.com/ihhub/penguinV/pull/159/files#r226814848).

You could add unit tests which are similar to bitmap operation (just save and load an image). Could you please as an extra verification step use load and save functions for gray-scale and RGB images to make that we do save/load in correct format?

ihhub:
As an extra small task: could you please revert added empty line made by me in `.gitignore`? I could't do this via web editor, it auto-sets empty line by default.

src/FileOperation/jpeg_file.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <jpeglib.h>
+#endif
+
+#include <setjmp>
+#include "jpeg_file.h"
+#include "../image_exception.h"
+
+namespace
+{
+    struct penguinv_err_mgr {
```
ihhub: Why do we need these structure and function in first place? Can't we just use our own C++ exception (`image_exception`)?

`setjmp` and `longjmp` are very dangerous and should not be used in C++ code

ngetahun: We have the option of using the default jpeg library error handler. But overriding it will allow us to throw more meaningful errors. I can remove this, but it would be nice to have in the future.

ngetahun:
> I made changes in your code but I haven't tested them. Additionally, we should address [this comment](https://github.com/ihhub/penguinV/pull/159/files#r226814848).
> 
> You could add unit tests which are similar to bitmap operation (just save and load an image). Could you please as an extra verification step use load and save functions for gray-scale and RGB images to make that we do save/load in correct format?

Sure thing.

src/FileOperation/jpeg_file.cpp
ihhub: Sorry, missed your reply ealier. We better do not use `setjmp` and `longjmp` so default handler would be fine. We need to only know whether we succeed to create an image or we failed.

ihhub: Okay, looks like you are right and jpeg library indeed uses such technique.

ihhub:
Keep your error handling code as it is. We will review them later if there would be a need.

ihhub:
Hi, @ngetahun , any news about this pull request?

ihhub:
Hi, @ngetahun are you still working on it? We're almost done this :)

ihhub:
@yannlabou could you please have a look into CMake file of this pull request? We need to add a support of libjpeg library (download it perhaps?).

yannlabou:
> @yannlabou could you please have a look into CMake file of this pull request? We need to add a support of libjpeg library (download it perhaps?).

Sure. Later today. Separate issue?

ihhub:
@yannlabou unit tests would fail for sure. I'll fix something for compilation pass at least.

yannlabou:
> @yannlabou unit tests would fail for sure. I'll fix something for compilation pass at least.

Because of Image declared const. I'm concerned about Appveyor not able to install `libjpeg`. We might need a separate issue for that.

yannlabou:
@ihhub Feel free to cancel https://ci.appveyor.com/project/ihhub/penguinv/builds/20279194

ihhub:
@yannlabou I think it would be better to create 2 separate issues: one to install and add support for **libjpeg**, another one - for **libpng**. Png code is already present in the library so it would be easier to do this first and then go for jpeg. I'll add unit tests for PNG soon. As of now you could just include `src/FileOperation/png.cpp` file into CMake file for the compilation.

If you look into `src/FileOperation/jpeg_file.cpp` file of this pull request you could notice that there is no Windows support as of now:
```cpp
#ifdef _WIN32
#else
#include <jpeglib.h>
#endif
```

The same applies for PNG. There are 2 ways to solve this:
1. Copy lipng/libjpeg library as third-party library into source code of the project. Disadvantage: extra code
2. Install it before CMake or ideally within CMake.

For PNG I will make changes to do not do anything on Windows so compilation would succeed at least with no body inside source file.

ihhub:
I'll create these issues.

ihhub:
Okay, Travis-CI built properly but as I expected unit tests fail.

yannlabou:
> @yannlabou I think it would be better to create 2 separate issues: one to install and add support for **libjpeg**, another one - for **libpng**. Png code is already present in the library so it would be easier to do this first and then go for jpeg. I'll add unit tests for PNG soon. As of now you could just include `src/FileOperation/png.cpp` file into CMake file for the compilation.
> 
> If you look into `src/FileOperation/jpeg_file.cpp` file of this pull request you could notice that there is no Windows support as of now:
> 
> ```c++
> #ifdef _WIN32
> #else
> #include <jpeglib.h>
> #endif
> ```
> 
> The same applies for PNG. There are 2 ways to solve this:
> 
> 1. Copy lipng/libjpeg library as third-party library into source code of the project. Disadvantage: extra code
> 2. Install it before CMake or ideally within CMake.
> 
> For PNG I will make changes to do not do anything on Windows so compilation would succeed at least with no body inside source file.

I strongly prefer option #2 (Let CMake download and install the lib) even though it will require lot of work on Windows.

ngetahun:
@ihhub @yannlabou The jpeg tests are throwing a segfault exception. I can't do memory debugging on my mac but I have an ubuntu vm and will fix it asap. As for the jpeg library, I think we can try using ```find_package``` else we can create a command to install and link it during build-time.

yannlabou:
> @ihhub @yannlabou The jpeg tests are throwing a segfault exception. I can't do memory debugging on my mac but I have an ubuntu vm and will fix it asap. As for the jpeg library, I think we can try using `find_package` utility from cmake, else we can create a command to install and link it during build-time.

`find_package` works if and only the lib is already available on the system. I'm not aware on any UNIX platforms without a libjpeg version. Sadly is Windows most of the time the problem.

ngetahun:
Yeah, exactly my point. The we only need ~3 files from the library and it's readily available on source forge. Or, we can just install the library in the ```thirdparty``` folder. The latter takes less effort and it's guaranteed to work.

ihhub:
> Yeah, exactly my point. The we only need ~3 files from the library and it's readily available on source forge. Or, we can just install the library in the `thirdparty` folder. The latter takes less effort and it's guaranteed to work.

Could you please point the location of these files? A few files are not a big problem to be a part of the code.

ihhub:
> > @ihhub @yannlabou The jpeg tests are throwing a segfault exception. I can't do memory debugging on my mac but I have an ubuntu vm and will fix it asap. As for the jpeg library, I think we can try using `find_package` utility from cmake, else we can create a command to install and link it during build-time.
> 
> `find_package` works if and only the lib is already available on the system. I'm not aware on any UNIX platforms without a libjpeg version. Sadly is Windows most of the time the problem.

Windows is the biggest problem. From another side we cannot force people to search the internet for available libraries. It's not user friendly :)

What we could do: if no library found we don't compile it so project won't have a support of such feature.

yannlabou:
> > > @ihhub @yannlabou The jpeg tests are throwing a segfault exception. I can't do memory debugging on my mac but I have an ubuntu vm and will fix it asap. As for the jpeg library, I think we can try using `find_package` utility from cmake, else we can create a command to install and link it during build-time.
> > 
> > 
> > `find_package` works if and only the lib is already available on the system. I'm not aware on any UNIX platforms without a libjpeg version. Sadly is Windows most of the time the problem.
> 
> Windows is the biggest problem. From another side we cannot force people to search the internet for available libraries. It's not user friendly :)
> 
> What we could do: if no library found we don't compile it so project won't have a support of such feature.

Be patient :). The libraries are available. We just need to integrate them in our build toolchain.

ihhub:
> @ihhub @yannlabou The jpeg tests are throwing a segfault exception. I can't do memory debugging on my mac but I have an ubuntu vm and will fix it asap. As for the jpeg library, I think we can try using `find_package` utility from cmake, else we can create a command to install and link it during build-time.

It's not only the issue for the build. I'll mark places which require a change once I reach a PC :)

src/FileOperation/jpeg_file.cpp
```diff
@@ -0,0 +1,124 @@
...
+        jpeg_create_compress(&cInfo);
+        jpeg_stdio_dest(&cInfo, file);
+
+        cInfo.image_width = image.width();
+        cInfo.image_height = image.height();
+        cInfo.input_components = image.colorCount();
+        cInfo.in_color_space = JCS_RGB; // use RGB color space
+
+        jpeg_set_defaults(&cInfo);
+        jpeg_start_compress(&cInfo, TRUE);
```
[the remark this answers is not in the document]
ngetahun: @ihhub the value is a macro from the jpeg library. It's used like that.

ihhub:
@ngetahun please make changes which were highlighted in comments.

ihhub:
@ngetahun I fixed almost all critical outstanding issues. Now the problem is in some internal crash of jpeg library due to possible incorrect usage of it. Could you please tell us whether you're planning to continue to work on this issue?

ihhub:
Closing this pull request as the author stopped responding for a long time.

Issue #123, opened 2018-10-08
Python wrapper for Image class


ihhub:
1. Add a python wrapper for `Image` class in which we are able to call member functions of `Image` class. Put new code in `src/python` folder.
2. Add a simple example of using new Python code and put it in `example/python/image`.

Relates to #103

MatthewMcGonagle:
I've played around a little bit with making C extensions of python, and I'm interested in try to make progress on this issue. Just to clarify, you want the extension to be built using the distutils python module? As in [the description in the python docs](https://docs.python.org/3/extending/building.html#building-c-and-c-extensions-with-distutils)?

ihhub:
Looks like yes. Ideally we should compile it in Travis-vi but this is another topic of discussion.

ngetahun:
I suggest using swig. I've been playing with this issue over the weekend. @MatthewMcGonagle do you have another suggestion?

MatthewMcGonagle:
@ngetahun I'm not familiar with swig. I was thinking about just doing it manually according to the [instructions in the pydocs](https://docs.python.org/3/extending/newtypes_tutorial.html), but swig sounds like it is easier.

ngetahun:
Ok cool. You can also look at the tensorflow repo on how they generate their python library [here](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python). All the `*.i` files are swig interface files.

ihhub:
If you are out of ideas where and what exactly to import please ask me. I'll guide you with this.

MatthewMcGonagle:
Sorry to not get back with a comment for a while. I was just getting more familiar with the swig as that seems to be the best solution. It looks like python's distutils can handle swig files directly; hopefully we just need to pass in swig_options = ['-c++'] to distutils.core.Extension(), [see section 2.3.2 of the docs for distutils](https://docs.python.org/3/distutils/setupscript.html).

MatthewMcGonagle:
Short question: when using swig, rvalue constructors such as `ImageTemplate(ImageTemplate &&)` come up as warning that they will be ignored since they make no sense in Python. Is it better to just not put them in the swig interface file (so there are no warnings, but the interface file is less complete for other interfacing)?

MatthewMcGonagle:
Another question: `Image::data()` should be overwritten in swig using `%extend` to return a list type compatible with python. Should it return a numpy array or just python lists?

ihhub:
[...]

PR #168, opened 2018-10-21
Python

MatthewMcGonagle:
Related to issue #123 Python Wrapper for Image Class.

Basic python wrappers for the Image class and a python implementation of the bitmap_operation example.

### DIFF
--- a/examples/python/bitmap_operation.py
+++ b/examples/python/bitmap_operation.py
@@ -1,0 +1,60 @@
+''' 
+Example of how to use python wrappers of penguinV to save a gray scale version of a bitmap image.
+'''
+
+# Module files for penguinV are in separate folder, so we use the sys module to add that location
+# to the places that python will look for the module.
+
+import sys
+sys.path.insert(0, '..\\..\\src\\python')
+
+# Now get the module.
+
+import penguinV 
+
+def saveGrayScaleCopy(filenameToOpen, filenameToSave):
+    ''' Saves a Gray Scale Thresholded version of a bitmap file.
+    
+    Parameters
+    ----------
+    filenameToOpen : String
+        Name of the bitmap to open.
+
+    filenameToSave : String
+        Name of the resulting bitmap.
+    '''
+
+    image = penguinV.Load(filenameToOpen)
+    
+    if image.empty():
+        raise penguinV.ImageException('penguinV.Load()', 'Problem loading file ' + filenameToOpen)
+
+    if image.colorCount() != penguinV.GRAY_SCALE:
+        image = penguinV.ConvertToGrayScale(image)
+
+    threshold_value = penguinV.GetThreshold( penguinV.Histogram(image) ) 
+    image = penguinV.Threshold(image, threshold_value)
+
+    penguinV.Save(filenameToSave, image)
+  
+########################
+### main execution 
+########################
+
+try:
+
+    saveGrayScaleCopy( filenameToOpen = '..\\_image\\mercury.bmp', 
+                       filenameToSave = 'result.bmp'
+                     ) 
+
+except penguinV.ImageException as err:
+
+        # This error occurs when there was trouble opening the file.
+
+        print(err)
+
+except:
+
+        # Something more serious has gone wrong.
+
+        print('Unknown Error.') 
--- a/examples/python/bitmap_operation.py
+++ b/examples/python/bitmap_operation.py
@@ -53,8 +53,8 @@ except penguinV.ImageException as err:
 
         print(err)
 
-except:
+except BaseException as err: 
 
         # Something more serious has gone wrong.
 
-        print('Unknown Error.') 
+        print('Unknown Error ', err) 
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -1,0 +1,26 @@
+# How to Create the Python Wrappers
+
+The wrappers for python are generated using SWIG ([website](http://www.swig.org/)), and for ease of use, the wrappers
+can also be generated with the help of the `distutils` python module. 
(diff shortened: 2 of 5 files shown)


MatthewMcGonagle:
It looks like CodeFactor doesn't like that the bitmap_operation.py version of bitmap_operation.cpp uses an `except` clause without any type of exception. However, the `.cpp` version does exactly this. What should change for the `.py` version?

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

MatthewMcGonagle:
I added a `BaseException` type for the generic exception catch in `bitmap_operation.py`. Now when I click on the "Details" of the CodeFactor complaint, it shows up as everything is okay. However, I don't know how to make it run again and show up success/fail on the pull-request thread.

examples/python/bitmap_operation.py
```diff
@@ -0,0 +1,60 @@
...
+        Name of the bitmap to open.
+
+    filenameToSave : String
+        Name of the resulting bitmap.
+    '''
+
+    image = penguinV.Load(filenameToOpen)
+    
+    if image.empty():
+        raise penguinV.ImageException('penguinV.Load()', 'Problem loading file ' + filenameToOpen)
```
ihhub: Why don't we just use `print` function and call normal return. In such situation we don't need an extra exception. What do you think?

Sorry for late reply, github was down for me almost whole day.

src/python/README.md
```diff
@@ -0,0 +1,26 @@
+# How to Create the Python Wrappers
```
ihhub: In think we need to put this file under `doc/python`.

ihhub:
All looks okay. I'm waiting for your reply about exception and then we could merge your changes.

ihhub:
I've tested on Windows. It runs smoothly but we need (in a future, not now) a proper to-do steps for compilation.

examples/python/bitmap_operation.py
MatthewMcGonagle: I was just trying to make the python version as close to the .cpp version as possible. Shouldn't we at least keep the `penguinV.ImageException` to showcase using the exceptions? If you think its unnecessary, then I'm totally fine with removing it.

I was also having a lot of trouble with github yesterday.

src/python/README.md
MatthewMcGonagle: Okay, but the instructions will only work from the `src/python` directory. If it is run from somewhere else, the `setup.py` file has the wrong relative path to find `penguinV.i`. Do we intend for users to run `python ..\..\src\python\setup.py build_ext --inplace` from the `doc/python` directory?

If so, I guess a quick workaround is to use relative paths such as `../../src/python/penguinV.i`, since `doc/python` happens to be the same number of levels deep as `src/python`. However, if we intend for `setup.py` to be run from the `doc/python` directory, maybe it would be better to have a separate `setup_doc.py` for that purpose, or to pass another command line argument to `setup.py`. What do you think?

examples/python/bitmap_operation.py
ihhub: Well, it's just an example. It should just work as `cpp` version but nobody demands to be an exact copy of it. I think we just go with what we have.

Github is still having problems.

src/python/README.md
ihhub: The purpose of `doc` directory is obviously to store a documentation. I missed some point that `cuda` and `opencl` have a README files inside their directories with a brief description in them. I think you initial location of the file was correct and I made a wrong suggestion about it as this file contains an information not related to API description. I'll put it back in a separate commit after merging your changes into the project.

The documentation ideally should be generated into `doc/python` directory and we could handle it in a separate issue.

ihhub:
@MatthewMcGonagle , you made a great work for the project. Thanks a lot for your contribution to the project!

ihhub:
I look forward for your future contribution to the project if you have a willingness to spend your time on it.

Issue #113, opened 2018-10-08
SIMD implementation of Accumulate() function


ihhub:
We do not have SIMD (SSE, AVX or NEON) implementation of **Accumulate()** function. for the person who is willing to add such code please take a look into an implementation of the function inside `image_function.cpp` and the structure which is used for SIMD code inside `image_function_simd.cpp`. Please do not forget to add unit test for newly created function(s) inside `unit_test_image_function.cpp` (it's just one line code via macros call).

0x72D0:
I will try to do this one( SSE, AVX and NEON ).

ihhub:
Sure. It should be very similar to `Sum` function in terms of instruction usage.

0x72D0:
in the AVX implementation do you accept the use of sse intrinsics?

ihhub:
`avx` has a full set of commands which are an equivalent to `sse`. Additionally, `avx` uses different registers on processor. So there is no reason to mix them.

0x72D0:
my problem is the following:
the accumulate function take the byte one by one in the memory and add them to the result vector, the byte is a uint8_t and the vector contains a uint32_t. so I was thinking using this intrinsic to convert the 8 bit integer to 32 bit integer: `__m256i _mm256_cvtepu8_epi32 (__m128i a)` but I need a __m128 register in order to use this function... so I was planning on using `__m128i _mm_loadu_si64 (void const* mem_addr)` which need sse in order to work

ihhub:
Won't the following fulfil your needs:
1. Read 256 bit variable (32 bytes)
2. Convert the variable into 2 variables of 16 bits (low and high) (`_mm256_unpacklo_epi8` and `_mm256_unpackhi_epi8`)
3. Convert into 4 variables of 32 bits (low and high from 16 bit) (`_mm256_unpacklo_epi16` and `_mm256_unpackhi_epi16`)
4. Add to original array all 4 variable (`_mm256_add_epi32`)

0x72D0:
from what I can see the function `__m256i _mm256_unpackhi_epi8 (__m256i a, __m256i b)` take two register and fuse them into one, so I can use a "zero" register to just keep the higher part, but still how can I zero extend them to 16 bit? Yeah the sum and the accumulate function look similar, but if I understand well in the Accumulate function I don't want the byte to add to other thing than the result so I can't use the `_mm256_add_epi16( dataLo, dataHi )` to cast them to 16 bit.

[...]

PR #169, opened 2018-10-22
add avx support for the accumulate function

0x72D0:
PR for #113. I add the avx support for the accumulate function, the only thing I found weird in my implementation is it seems to be slower on larger dataset:

screenshot_20181021_224430

I should implement the sse support in the following days.

### DIFF
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -62,10 +62,9 @@ namespace avx
         for( ; imageY != imageYEnd; imageY += rowSize, outY += rowSize ) {
             const simd * src    = reinterpret_cast <const simd*> (imageY);
             const simd * srcEnd = src + simdWidth;
-            const simd * dst_const = reinterpret_cast <const simd*> (outY);
-            simd       * dst  = reinterpret_cast <simd*> (outY);
+            simd       * dst    = reinterpret_cast <simd*> (outY);
 
-            for( ; src != srcEnd; ++src, dst+=4, dst_const+=4) {
+            for( ; src != srcEnd; ++src ) {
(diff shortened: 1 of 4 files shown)


ihhub:
Please test the code after my corrections. Regarding the performance tests: your code is slightly faster than CPU first of all and with my tiny changes it should be even faster. Please remember that compiler `could` use SIMD optimizations as well leading to almost identical code.

0x72D0:
wait there's something I don't understand... The reason why I put the dst_const variable is because without it the loadu function give me a segfault... so why there's is no segfault now even if you pass a non const variable to loadu?

ihhub:
`segfault` could happen because of reading a data which is out of bound for example. Passing a `non-const` argument to a function which requires a `const` argument automatically makes the argument `const`. Please verify on unit tests that everything works fine. If yes we could merge the code :)

0x72D0:
yes everythings work fine on my side

ihhub:
@0x72D0, thanks a lot again for your work for the project and I hope that you would continue your contribution!

ihhub:
There was a bug in the code which I fixed in [this commit](https://github.com/ihhub/penguinV/commit/d5a9ca02c467fdf1533269f639a9f96b3c7416fa)

0x72D0:
what was the bug exactly?

ihhub:
`outY` which is a pointer to `result` vector was incremented by `rowSize` (instead of `width`) which is bigger than `width` for situations of ROI (not full) image processing. That's why **form2** of the test had **segfault** (out of bounds).

0x72D0:
oh! okay I see, thanks for the precision :)

PR #170, opened 2018-10-22
Add OpenCL check in CMake and add cmake files for unit tests

yannlabou:
Adding CMake config for test/unit_tests/opencl

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ addons:
   apt:
     packages:
       - libpng-dev
+      - ocl-icd-opencl-dev
+      - opencl-headers
 
 matrix:
   include:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,8 +29,8 @@ build_script:
   - cmd: cd C:\projects\penguinv
   - cmd: md build
   - cmd: cd build
-  - cmd: if "%platform%"=="x64" cmake -G "Visual Studio 14 2015 Win64" -DCMAKE_BUILD_TYPE=%configuration% ..
-  - cmd: if "%platform%"=="x86" cmake -G "Visual Studio 14 2015" -DCMAKE_BUILD_TYPE=%configuration% ..
+  - cmd: if "%platform%"=="x64" cmake -G "Visual Studio 14 2015 Win64" -DPENGUINV_BUILD_UNIT_TEST_OPENCL=OFF ..
+  - cmd: if "%platform%"=="x86" cmake -G "Visual Studio 14 2015" -DPENGUINV_BUILD_UNIT_TEST_OPENCL=OFF ..
   - cmd: cmake --build . --config %configuration%
 
 test_script:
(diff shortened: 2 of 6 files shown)


ihhub:
@yannlabou , I restarted tests on Travis-CI after the fix for AVX failure in master branch. The compilation on **Linux** fails because the script couldn't find `OpenCL` module/directory. There are 2 ways to fix it:
1. Do not compile (and execute) OpenCL unit tests if OpenCL module is not found.
2. Try to find CUDA module and then use OpenCL code there. CUDA usually ships with OpenCL module.

yannlabou:
I will push a mix of option 1 and 2.

ihhub:
I fixed OpenCL compilation on Linux so Travis-CI does not fail anymore. I need some time to figure out why AppVeyor (Windows) fails.

ihhub:
**Appveyor** fails to raised exception in OpeCL unit tests. Could you please modify `main` function of `penguinV/test/unit_tests/opencl/unit_tests_opencl.cpp` file?

```
int main()
{
    try
    {
    ... existing code shifted by 4 pixels to right
    }
    catch( const imageException & ex ) {
        std::cout << "Exception " << ex.what() << " raised during OpenCL unit tests." << std::endl;
        return 0;
    }
    catch( ... ) {
        std::cout << "Unknown exception raised during OpeCL unit tests." << std::endl;
        return 0;
    }
```

You could refer to `penguinV/examples/bitmap_operation/example_bitmap_operation.cpp` file to understand what I mean.
This approach might help to fix OpenCL run failure and identify the source of failure as well.

yannlabou:
@ihhub Thank you! What was the issue on Linux? I can't see the build log. I will take care of  `penguinV/test/unit_tests/opencl/unit_tests_opencl.cpp` later today.

ihhub:
`#include <memory>` was missing which by default on Linux leads to compilation failure for `std::shared_ptr` while for Windows and MacOS it was fine.

yannlabou:
Sure. `std::shared` is defined in `memory` header. I'm wondering why my clang on OS X was happy with that. Thank you anyway :).

ihhub:
I changed the description of the issue and pull request and we don't have performance tests for OpenCL. I will create another issue for performance tests CMake file when tests would be ready.

ihhub:
No luck in exception handling. The change didn't fix the issue. 2 ways to solve it:
1. do not run OpenCL tests on Windows
2. Copy (which one) OpenCL dll to the directory where we create ***.exe** file. Not sure how to find our which file, need to dig the Internet for this :(

yannlabou:
> No luck in exception handling. The change didn't fix the issue. 2 ways to solve it:
> 
> 1. do not run OpenCL tests on Windows
It's doable
> 2. Copy (which one) OpenCL dll to the directory where we create ***.exe** file. Not sure how to find our which file, need to dig the Internet for this :(
If you're 100% sure that the exception is caused by some missing OpenCL DLLs, I can add an [CMake install](https://cmake.org/cmake/help/v3.9/command/install.html#installing-files) directive for the unit_tests_opencl target.

ihhub:
@yannlabou , thanks a lot for your continuous contribution to the project!

PR #176, opened 2018-10-23
add sse support for the accumulate function

0x72D0:
the performance for the sse support:

screenshot_20181022_211751

### DIFF
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -1345,7 +1345,7 @@ if ( simdType == neon_function ) { \
         const uint32_t simdSize = getSimdSize( simdType );
         const uint8_t colorCount = image.colorCount();
 
-        if( (simdType != avx_function) || ((width * height * colorCount) < simdSize) ) {
+        if( (simdType == cpu_function) || (simdType == neon_function) || ((width * height * colorCount) < simdSize) ) {
             Image_Function::Accumulate(image, x, y, width, height, result);
             return;
         }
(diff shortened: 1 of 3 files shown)


src/image_function_simd.cpp
```diff
@@ -1326,6 +1369,7 @@ if ( simdType == neon_function ) { \
         const uint32_t nonSimdWidth = width - totalSimdWidth;
 
         AVX_CODE( avx::Accumulate( rowSize, imageY, imageYEnd, outY, simdWidth, totalSimdWidth, nonSimdWidth ); )
+        SSE_CODE( sse::Accumulate( rowSize, imageY, imageYEnd, outY, simdWidth, totalSimdWidth, nonSimdWidth ); )
```
ihhub: You forgot to change line **1348** and your code never reaches this line. That line should be `if( (simdType == cpu_function) || (simdType == neon_function) || ((width * height * colorCount) < simdSize) ) {`

0x72D0:
this is the cpu I use for the test, a x86_64:

screenshot_20181022_213121

ihhub:
@0x72D0 if you facing a consistent crash with blob detection code could you please create a separate issue for this describing the problem and adding more information? Ideally if you run only blob detection in debug mode so you could define the place of crash.

0x72D0:
ok I'm gonna add a new issue

src/image_function_simd.cpp
0x72D0: ok it should be fix now

ihhub:
@0x72D0, thanks again for you contribution to the project!

Issue #177, opened 2018-10-23
segfault error in the blob detection code execution


0x72D0:
when I run the unit_tests, I get a segfault at the blob_detection tests:

![screenshot_20181022_220205](https://user-images.githubusercontent.com/16087328/47330468-5829f680-d646-11e8-95df-ef0d8f349a09.png)

when I run gdb, it seems a blobInfo destruction problem, here's the backtrace:

![screenshot_20181022_220609](https://user-images.githubusercontent.com/16087328/47330589-c79fe600-d646-11e8-8aa8-5edbe57675f3.png)

and here's the cpu I use:

![screenshot_20181022_213121](https://user-images.githubusercontent.com/16087328/47330621-dc7c7980-d646-11e8-93a9-3925e361497b.png)

Also the problem is really recent, one week ago I was able to run the blob_detection function without a problem.

ihhub:
Do you have segfault before [this commit](https://github.com/ihhub/penguinV/commit/5648fef399fac6be5a9d0839617100cffc83f42b)?

0x72D0:
I don't really know... but if it was a modification 2 day ago, it might be the cause

0x72D0:
wait I will try to checkout to before this commit

0x72D0:
ok when I checkout to the commit just before this one, blob_detection runs ok.

0x72D0:
ok when I go to the commit you gave, it just won't compile because of a bad path to math.h, but when I checkout to that [commit](https://github.com/ihhub/penguinV/commit/864905d140f36a46733331798576c052262e72e5), I have a segfault with blob_detection

ihhub:
Thanks. That's useful.

ihhub:
Could you please do following changes `math_base.h` file and then test again:
1. At the beginning of the file after inclusion of `cmath` add this code:
> namespace
> {
>     const double epsilonDouble = 1e-10;
> 
>     template <typename _Type>
>     bool isEqual( const _Type & value1, const _Type & value2 )
>     {
>         return ( value1 == value2 );
>     }
> 
>     template <>
>     bool isEqual<double>( const double & value1, const double & value2 )
>     {
>         return fabs( value1 - value2 ) < epsilonDouble;
>     }
> }

In `virtual bool operator == ( const PointBase2D & point ) const` body put this code:
`return isEqual(x, point.x) && isEqual(y, point.y);`

and in `virtual bool operator == ( const PointBase3D & point ) const` body put this code:
[...]

PR #181, opened 2018-10-23
Fix segfault for blob detection code execution on Linux

0x72D0:
PR for #177


ihhub:
@0x72D0 , thanks a lot for your contribution to the project!

Issue #179, opened 2018-10-23
Modify unit tests for Bitmap operation


ihhub:
Current unit test (in `test/unit_tests/unit_test_bitmap.cpp` file) for Bitmap does not cover all situations. What we need is:
1. Rename `LoadSaveGrayScaleImage` into `WhiteGrayScaleImage`. Change filename from "white.bmp" to "bitmap.bmp". Set `PenguinV_Image::Image original` and `PenguinV_Image::Image loaded`
2. Add a new function called `BlackGrayScaleImage` withe the same functionality but to save/load black image.
3. Add a new function called `RandomRGBImage` which obviously does the same as above but generate random RGB image (please refer to `test/unit_tests/unit_test_helper.cpp`). Use `memcmp` to compare each line of images (refer to `src/image_function.cpp` **IsEqual** function as an example.
4. Instead of macros `ADD_TEST( framework, bitmap_operation::LoadSaveGrayScaleImage );` use a direct call `framework.add(WhiteGrayScaleImage, "load and save white gray-scale image"` to have human readable output of tests.

lrm25:
I'll take this one

ihhub:
Sure.

PR #188, opened 2018-10-25
Issue #179:  modify unit tests for bitmap operation

lrm25:


ihhub:
I have to check why RGB image tests fails.

lrm25:
Just saw the AppVeyor message, and looked more closely at the image on my machine.  It isn't being saved properly.  Hang on.

lrm25:
In fact, there's no random RGB image generator in the unit test helper.  I can add that too.

ihhub:
@lrm25 , thank you very much for your contribution to the project!

Issue #187, opened 2018-10-25
Use single buffer allocation in Dilate() function


ihhub:
We have `Dilate` function in unnamed namespace of `src/image_function.cpp` files. There are 2 places where we allocate 2 arrays:
```cpp
uint8_t ** startPos = new uint8_t *[width];
uint8_t ** endPos   = new uint8_t *[width];
```

and 2 places where we deallocate memory:
```cpp
delete[] startPos;
delete[] endPos;
```

Better solution is to allocate a single buffer but in 2 times bigger and assign a pointer to another buffer from first with **shift** in address. Deallocation would be done only once in this case.

Additionally remove this chunk of code
```cpp
if( dilationX == 0u && dilationY == 0u )
    return;
```

PR #189, opened 2018-10-25
Use sigle buffer allocation/deallocation

ya1gaurav:
Fix for issue reported in https://github.com/ihhub/penguinV/issues/187

### DIFF
--- a/src/image_function.cpp
+++ b/src/image_function.cpp
@@ -19,7 +19,8 @@ namespace
         if( dilationX > 0u ) {
             const int32_t dilateX = static_cast<int32_t>(dilationX);
 
-            uint8_t ** imageWPos = new uint8_t *[width*2 + 1];
+            uint8_t ** startPos = new uint8_t *[2 * width];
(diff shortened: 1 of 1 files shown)


src/image_function.cpp
```diff
@@ -21,15 +19,15 @@ namespace
         if( dilationX > 0u ) {
             const int32_t dilateX = static_cast<int32_t>(dilationX);
 
-            uint8_t ** startPos = new uint8_t *[width];
```
ihhub: `uint8_t ** startPos = new uint8_t *[2 * width];`

src/image_function.cpp
```diff
@@ -21,15 +19,15 @@ namespace
...
-            uint8_t ** endPos   = new uint8_t *[width];
```
ihhub: `uint8_t ** endPos   = startPos + width;`

ihhub:
Thank you very much for contribution to the project!

Issue #183, opened 2018-10-24
Remove generated bitmap file after unit test


ihhub:
We have unit test for **Bitmap** in `penguinV/test/unit_tests/unit_test_bitmap.cpp` where we generate (save) a bitmap file and then load it but we did not remove the file (cleanup). It's good to do this.

lrm25:
I can take this one too.  Do you want an error message, or do you want it to just quietly fail if it does?

ihhub:
Quietly delete a file after loading it inside each test. No need an error message as it's just garbage cleaner code.

ihhub:
File deletion function put please inside `unit_test_helper`.

lrm25:
Do you need a separate function?  File deletion without error handling is just "remove(<path>)".  If so, I can add it though.

ihhub:
Sorry, I think you are right. We don't need a separate function as this is just one line function call.

PR #193, opened 2018-10-26
Issue #183:  remove generated bmp file after unit test

lrm25:


test/unit_tests/unit_test_bitmap.cpp
```diff
@@ -6,34 +6,39 @@ namespace bitmap_operation
 {
```
ihhub: Add `#include <stdio.h>`

test/unit_tests/unit_test_bitmap.cpp
```diff
@@ -6,34 +6,39 @@ namespace bitmap_operation
...
     bool WhiteGrayScaleImage()
     {
+        bool passed = true;
```
ihhub: Remove this variable and all code related to it.

test/unit_tests/unit_test_bitmap.cpp
```diff
@@ -6,34 +6,39 @@ namespace bitmap_operation
...
         const PenguinV_Image::Image original = Unit_Test::whiteImage();
         Bitmap_Operation::Save( "bitmap.bmp", original );
 
         const PenguinV_Image::Image loaded = Bitmap_Operation::Load( "bitmap.bmp" );
```
ihhub: Remove file here after loading.

test/unit_tests/unit_test_bitmap.cpp
lrm25: I replaced the return value with a variable so that the image would be deleted, regardless of whether or not the test passed or failed.  What do you want me to do instead?

test/unit_tests/unit_test_bitmap.cpp
lrm25: Got it

ihhub:
@lrm25 , thanks a lot again for the contribution to the project!

PR #197, opened 2018-10-27
Some fixes for python example in Linux.

dianarg:
- Use os.path.join to support cross-platform paths.
- Windows supports forward slash in include paths, so changed paths to be compatible with Linux.


ihhub:
@dianarg , thanks a lot for the fix and for the contribution to the project!

PR #198, opened 2018-10-27
Run cppclean on src, test, examples only.

### LINKED ISSUE #196: Run static analyser only on certain directories in Travis-CI job via bash script
We have a job in Travis-CI job to run static analysis on our own code (`script/travis/run_script.sh`):
```bash
git clone https://github.com/myint/cppclean
./cppclean/cppclean .
```

Such code run full analysis on **cppclean** itself too which we don't need to do. We want to run cppclean only on `src`, `test` and `examples` directories.

dianarg:
- Fixes #196.


ihhub:
Thank you very much for the contribution to the project!

PR #200, opened 2018-10-28
Add Qt Support for Travis CI

yannlabou:

PR #206, opened 2018-10-28
Travis CI must fail on build failure

yannlabou:


ihhub:
@yannlabou , I have an idea: what if we move back the logic from `script/travis/run_script.sh` file into `.travis`. I see in your [another pull request](https://github.com/ihhub/penguinV/pull/200/files) that actually we **can** use environmental variable within `.travis` file. If we move back everything then we don't need to have `pipeline` fix. What do you think?

ihhub:
Oh, now we have the same changes in both pull request. Could you please create a separate branch for this issue from the master branch of the project then you would have 2 separate copies to work on?

yannlabou:
I think I merged the two issue-branches inadvertently. :(

ihhub:
That's okay :) No blame on you.

PR #218, opened 2018-10-30
Removed FunctionRegistrator inheritance from penguinV::FunctionTable

### LINKED ISSUE #217: Do not inherit FunctionRegistrator from penguinV::FunctionTable
We have 4 independent structures of `FunctionRegistrator` in these files:
```
src/image_function.cpp
src/image_function_simd.cpp
src/opencl/image_function_opencl.cpp
src/cuda/image_function_cuda.cu
```

`FunctionRegistrator` structure is inherited from `penguinV::FunctionTable`. In constructor we initialise table/structure and then we make a call of `penguinV::registerFunctionTable` with `this` pointer which is wrong and could lead to uncertain situations.

We need to do **not** inherit from `penguinV::FunctionTable` but has a member of structure (call it `table` and then pass this member into the function.

This code might lead to random segfault on AppVeyor.

kgorecki:
I hope that this is what you meant. Resolves #217

### DIFF
--- a/src/cuda/image_function_cuda.cu
+++ b/src/cuda/image_function_cuda.cu
@@ -9,30 +9,32 @@
 
 namespace
 {
-    struct FunctionRegistrator : public penguinV::FunctionTable
+    struct FunctionRegistrator
     {
+        penguinV::FunctionTable table;
+
         FunctionRegistrator()
         {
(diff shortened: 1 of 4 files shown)


ihhub:
Yep, it's what I meant. Let's wait for AppVeyor compilation completion.

ihhub:
@kgorecki , thanks a lot for your continuous contribution to the project!

PR #221, opened 2018-10-30
Add unit tests for FFT

ihhub:
Added unit tests for FFT CPU based classes.

### DIFF
--- a/src/thirdparty/kissfft/kiss_fft.c
+++ b/src/thirdparty/kissfft/kiss_fft.c
@@ -223,7 +223,7 @@ static void kf_bfly_generic(
             int twidx=0;
             Fout[ k ] = scratch[0];
             for (q=1;q<p;++q ) {
-                twidx += fstride * k;
+                twidx += (int)(fstride * k);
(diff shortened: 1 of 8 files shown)


ihhub:
Hi, @yannlabou I think the option [`/WX`](https://msdn.microsoft.com/en-us/library/ms235592.aspx) for VIsual Studio projects is wrong. It causes compilation error.

yannlabou:
> Hi, @yannlabou I think the option [`/WX`](https://msdn.microsoft.com/en-us/library/ms235592.aspx) for VIsual Studio projects is wrong. It causes compilation error.

Why? There is nothing inherently bad in treating linker warnings as errors. Do you have a build log for me?

ihhub:
Check [this log](https://ci.appveyor.com/project/ihhub/penguinv/builds/19907718/job/klstu3u6epp7ovym). It was a simple casting warning. I fixed it by explicitly add casting. Let's monitor for another build if any new code would cause the problem again.

ihhub:
Line 226 was `twidx += fstride * k;` which is totally fine and it doesn't relate to linker.

PR #232, opened 2018-11-01
Adding valgrind environment on Travis CI

yannlabou:


yannlabou:
> Hi, @yannlabou , let me know when I should review changes. Currently we have 2 problems:
> 
> 1. Tests fail due to timeout
> 2. Travis takes wooping 58 minutes
> 
> I have an idea to allow to set iteration count for loops inside unit tests via application command parameters. Now we have a fixed value 1001. Valgrind makes the application pretty slow so setting this value, let's say, to 10 for now fixes two above issues. What do you think?

Sure! Let's give it a try. I'm building right now in release mode to check if we are still running out of time.

yannlabou:
@ihhub How to reduce Valgrind iteration count?

ihhub:
It's not valgrind. Search for runCount function inside test/unit_tests/unit_test_helper.cpp file

yannlabou:
> It's not valgrind. Search for runCount function inside test/unit_tests/unit_test_helper.cpp file

Changing the iteration count to `101u` fixed the time out issue. You can review the changes.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
ihhub: Do we need this file?

test/unit_tests/unit_test_helper.cpp
```diff
@@ -328,7 +328,7 @@ namespace Unit_Test
 
     uint32_t runCount()
     {
-        return 1001u; // some magic number for loop. Higher value = higher chance to verify all possible situations
+        return 101u; // some magic number for loop. Higher value = higher chance to verify all possible situations
```
ihhub: Please revert this change and use command line input parameter for the unit test compiled application.

.travis.yml
```diff
@@ -41,3 +57,7 @@ script:
       cd ../test/unit_tests/cuda;
       make;
     fi
+
+after_script:
+  - sleep 30s
```
ihhub: Definitely this is not 100% working and correct solution and it wouldn't work always. I have multiple suggestions:
1. Put `valgrind` code execution to a script. It might be (no guarantee) that file would be saved before exiting the script as outstanding task. After the script we could easily display results
2. If above suggestion doesn't work try to use loop inside script, something like this (pseudo-code):
```bash
while( file is not created or file cannot be opened with write permission) then
    sleep 1 second
end
```
3. If non of solution work try to search for the viewers/text editors which could read file until writing to it would be finished

ihhub:
Hi @yannlabou if you want to see valgrind warnings try to revert [this change](https://github.com/ihhub/penguinV/commit/842dd4277cc22c6cf5bb07cf5308bbdf1c589cae)

ihhub:
I think we're missing valgrind call itself. I cannot find the place where we call it. We should call valgrind with these commands: `valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes --error-limit=no --log-file=some path to store`

yannlabou:
> I think we're missing valgrind call itself. I cannot find the place where we call it. We should call valgrind with these commands: `valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes --error-limit=no --log-file=some path to store`

This is exactly what `ctest --test-command unit_tests 101 --extra-verbose -D ExperimentalMemCheck -j2` is doing or at least the `-D ExperimentalMemCheck` part.

ihhub:
Hi, @yannlabou What's missing in the pull request or is everything done? I see that we still have very short output from valgrind. For example I had few hundred of warnings like this:
```bash
==29814== Conditional jump or move depends on uninitialised value(s)
==29814==    at 0x4C35E20: bcmp (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==29814==    by 0x168872: equalData<unsigned char> (unit_test_helper.h:53)
==29814==    by 0x168872: bool template_image::_AssignmentOperator<unsigned char>() (unit_test_image_buffer.cpp:81)
==29814==    by 0x15ABC1: UnitTestFramework::run() const (unit_test_framework.cpp:22)
==29814==    by 0x10CD88: main (unit_tests.cpp:28)
==29814==  Uninitialised value was created by a heap allocation
==29814==    at 0x4C3089F: operator new[](unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==29814==    by 0x1117D6: _allocate (image_buffer.h:271)
==29814==    by 0x1117D6: resize (image_buffer.h:82)
==29814==    by 0x1117D6: PenguinV_Image::ImageTemplate<unsigned char>::ImageTemplate(unsigned int, unsigned int, unsigned char, unsigned char) (image_buffer.h:29)
==29814==    by 0x168689: bool template_image::_AssignmentOperator<unsigned char>() (unit_test_image_buffer.cpp:76)
==29814==    by 0x15ABC1: UnitTestFramework::run() const (unit_test_framework.cpp:22)
==29814==    by 0x10CD88: main (unit_tests.cpp:28)
```

yannlabou:
> Hi, @yannlabou What's missing in the pull request or is everything done? I see that we still have very short output from valgrind.
> 

This [line](https://github.com/ihhub/penguinV/blob/c9a156c56a305b2b98c248b357e2fe83581f3503/.travis.yml#L24) is not getting executed and I still don't know why. I have the feeling that Travis is not running the commands `after_script`, `after_success`, and `after_failure` at all. Valgrind is definitely working.

ihhub:
> > Hi, @yannlabou What's missing in the pull request or is everything done? I see that we still have very short output from valgrind.
> 
> This [line](https://github.com/ihhub/penguinV/blob/c9a156c56a305b2b98c248b357e2fe83581f3503/.travis.yml#L24) is not getting executed and I still don't know why. I have the feeling that Travis is not running the commands `after_script`, `after_success`, and `after_failure` at all. Valgrind is definitely working.

Let's print something for now in every script to see that each is executed after all. Plus let's try (if first idea doesn't work) to move script call command on a separate line like:
```bash
after_success
  - script/travis/print_valgrind_log.sh
```

script/travis/print_valgrind_log.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+if [[ "${DYNAMIC_ANALYSIS}" == "ON" ]]; then
```
ihhub: Shouldn't it be like `if [[ "$DYNAMIC_ANALYSIS" == "ON" ]]; then` without `{}`?

yannlabou: 25 commits on this PR :). Believe me, I tried every single combination. Also every combination of `after_success` and alike. Let's wait 5 minutes and see if my last change helped.

ihhub: You could set the number of iterations to 10 (it would be even faster). I can cancel previous compilations if you push multiple commits.

yannlabou: Still no success at printing valgrind log.

ihhub:
Let's do these steps:
1. Set iteration count from *101* to *10*
2. Return back `after_success: script/travis/print_valgrind_log.sh`
3. In `print_valgrind_log.sh` remove `if [[ "${DYNAMIC_ANALYSIS}" == "ON" ]]; then` check. We know that we execute only from one job
4. Print something in `print_valgrind_log.sh` file like `echo print_valgrind_log.sh is being executed`

Then we would know is it a problem of non executing `print_valgrind_log.sh` or valgrind log is not valid.

yannlabou:
Could you cancel [Build #418](https://travis-ci.org/ihhub/penguinV/builds/450922309)? Editing from web triggers a new build for each file commit.

ihhub:
So `after_success` is not called at all. Sorry to ask but have you tried to call this script for `after_script` and for `after_failure` (logically call script in any way)?

yannlabou:
> So `after_success` is not called at all. Sorry to ask but have you tried to call this script for `after_script` and for `after_failure` (logically call script in any way)?

My first attempt was calling it at `after_script`. No success.

ihhub:
I suspect that `after_script:` would not succeed because `script` locates outside `matrix`. In this case we do not (some kind of ) script for our job. Putting `after_script` after `script` itself at the very end of the file might solve the problem.

yannlabou:
> I suspect that `after_script:` would not succeed because `script` locates outside `matrix`. In this case we do not (some kind of ) run script for the job. Putting `after_script` after `script` itself at the very end of the file might solve the problem.

I did that :). That's exactly why I had the check `if [[ "${DYNAMIC_ANALYSIS}" == "ON" ]]`. I will give it another try.

ihhub:
> > I suspect that `after_script:` would not succeed because `script` locates outside `matrix`. In this case we do not (some kind of ) run script for the job. Putting `after_script` after `script` itself at the very end of the file might solve the problem.
> 
> I did that :). That's exactly why I had the check `if [[ "${DYNAMIC_ANALYSIS}" == "ON" ]]`. I will give it another try.

Sorry for bugging you with suggestions as I haven't seen full changes in all your commits :(

yannlabou:
> > > I suspect that `after_script:` would not succeed because `script` locates outside `matrix`. In this case we do not (some kind of ) run script for the job. Putting `after_script` after `script` itself at the very end of the file might solve the problem.
> > 
> > 
> > I did that :). That's exactly why I had the check `if [[ "${DYNAMIC_ANALYSIS}" == "ON" ]]`. I will give it another try.
> 
> Sorry for bugging you with suggestions as I haven't seen full changes in all your commits :(

No big deal. It's unfortunately hard to review as I tried a lot. Valgrind itself was actually easy to setup :)

ihhub:
Okay. Let's try one last time for compilation. If it doesn't work let's polish changes what we have and make the commit **without** printing log file generated from valgrind. At least we can see the number of defects which is still useful after all.

For commit we stay with iteration count **10** but **only for debug version**. We need to remove `echo` printing and set `if` guard back. We could return to this issue later after finding a proper solution :)

yannlabou:
> Okay. Let's try one last time for compilation. If it doesn't work let's polish changes what we have and make the commit **without** printing log file generated from valgrind. At least we can see the number of defects which is still useful after all.
> 
> For commit we stay with iteration count **10** but **only for debug version**. We need to remove `echo` printing and set `if` guard back. We could return to this issue later after finding a proper solution :)

I'll implement your comments in a couple of hours. I have to leave the office ;)

.travis.yml
```diff
@@ -19,19 +19,35 @@ matrix:
...
       exit 0;
     fi
+  - |
+    if [[ "$DYNAMIC_ANALYSIS" == "ON" ]]; then
+      mkdir -p build && cd build
+      cmake .. ${CMAKE_OPTIONS}
+      cmake --build . --target unit_tests -- -j2
+      ctest -D ExperimentalBuild -j2
+      ctest -E perf_test --extra-verbose -T MemCheck -j2
+      exit 0
```
ihhub: It could be that this line causes us to dot not execute `after_script`. Try to remove it and see what happens.

.travis.yml
```diff
@@ -19,19 +19,35 @@ matrix:
...
 script:
-  - if [[ "$STATIC_ANALYSIS" == "ON" ]]; then
+  - |
+    if [[ "$STATIC_ANALYSIS" == "ON" ]]; then
       export EXCLUDE_PATTERN_1=" not found in expected header ";
       export EXCLUDE_PATTERN_2=" does not need to be \#included; use a forward declaration instead";
       export EXCLUDE_PATTERN_3=" not found in any directly \#included header";
       git clone https://github.com/myint/cppclean;
       ./cppclean/cppclean src test examples | grep -v "$EXCLUDE_PATTERN_1" | grep -v "$EXCLUDE_PATTERN_2" | grep -v "$EXCLUDE_PATTERN_3";
       exit 0;
```
ihhub: See for static analysis we don't execute `after_script` too and we have the same command here.

yannlabou: You're fully right. But without `exit`, the subsequent commands will be executed. I must come up with a better logic.

ihhub: You could try something like this:
```bash
  - mkdir build
  - cd build
  - cmake .. ${CMAKE_OPTIONS}
  - |
    if [[ "$DYNAMIC_ANALYSIS" == "ON" ]]; then
      mkdir -p build && cd build
      cmake .. ${CMAKE_OPTIONS}
      cmake --build . --target unit_tests -- -j2
      ctest -D ExperimentalBuild -j2
      ctest -E perf_test --extra-verbose -T MemCheck -j2
    else
      cmake --build .
       ctest -E perf_test --extra-verbose
    fi
```

`CUDA` doesn't need changes

.travis.yml
```diff
@@ -41,3 +56,6 @@ script:
       cd ../test/unit_tests/cuda;
       make;
     fi
+
+after_script:
+  - script/travis/print_valgrind_log.sh
```
ihhub: Here should be `../script/travis/print_valgrind_log.sh` as we execute the script from `build` directory.

yannlabou:
Printing valgrind output finally works.

ihhub:
Yay! That's cool! Let's prepare the code for commit.

ihhub:
@yannlabou , thanks a lot for this titanic work!

PR #233, opened 2018-11-02
Include BinaryDilate and BinaryErode functions in unit tests

lrm25:
…ests

### DIFF
--- a/test/unit_tests/unit_test_image_function.cpp
+++ b/test/unit_tests/unit_test_image_function.cpp
@@ -353,13 +353,18 @@ namespace Function_Template
 
     bool form1_BinaryDilate(BinaryDilateForm1 BinaryDilate)
     {
-        std::vector <uint8_t> v (100, 255u);
-        v.push_back(0u);
-        PenguinV_Image::Image image = randomImage(v);
+        std::vector< uint8_t > fillData( randomValue<uint32_t>(20, 200), 255u );
+        fillData.push_back(0u);
 
-        BinaryDilate(image, randomValue<uint32_t>(1, image.width()), randomValue<uint32_t>(1, image.height()));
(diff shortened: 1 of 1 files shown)


test/unit_tests/unit_test_image_function.cpp
```diff
@@ -351,6 +351,68 @@ namespace Function_Template
...
+
+    bool form2_BinaryDilate(BinaryDilateForm2 BinaryDilate)
+    {
+        std::vector <uint8_t> v (100, 255u);
+        v.push_back(0u);
+        PenguinV_Image::Image image = randomImage(v);
+
+        uint32_t roiX, roiY, roiWidth, roiHeight;
+        generateRoi ( image, roiX, roiY, roiWidth, roiHeight );
+        uint8_t result = !verifyImage (image, roiX, roiY, roiWidth, roiHeight, 0u) ? 255u : 0u;
```
ihhub: Could you please explain why do we need to do this?

ihhub:
I've changed full ROI functions (through web editor). Please have a look into changes :)

test/unit_tests/unit_test_image_function.cpp
lrm25: verifyImage will return white if the random image region can be dilated, and if the region is fully black (and can't be dilated), black will be returned.  I was comparing this to the value returned from BinaryDilate to see if the dilation results matched what I predicted.  If it's not right, let me know why, or if there's an easier or better way, let me know.

I'll pull that makefile out and put it in a new PR.

ihhub: Okay. I got the idea. I corrected the logic of functions to be more obvious for other people.

ihhub:
@lrm25 , thank you very much for your continuous support for the project! I hope you would continue to do the same :)

PR #245, opened 2018-11-06
Disable valgrind unit test by default

yannlabou:


ihhub:
@yannlabou thank you very much for your continuous contribution to the library!

Issue #116, opened 2018-10-08
SIMD implementation of ProjectionProfile() function


ihhub:
We do not have SIMD (SSE, AVX or NEON) implementation of **ProjectionProfile()** function. For the person who is willing to add such code please take a look into an implementation of the function inside `src/image_function.cpp` and the structure which is used for SIMD code inside `src/image_function_simd.cpp`. Please do not forget to add unit test for newly created function(s) inside `test/unit_tests/unit_test_image_function.cpp` (it's just one line code via macros call) as well as to add performance tests (`tests/performance_tests/performance_test_image_function.`).

For projection on X axis we need to have similar logic which *Accumulate* function does while for Y axis projection we must have a logic similar to *Sum* function.

ihhub:
Hi, @0x72D0. This issue is very similar to #113 . Would you like to have a look?

0x72D0:
Yeah it would be a pleasure for me to try this one, but I'm a little bit occupied this week, so maybe I can start it this weekend :)

ihhub:
No rush, take your time. Thanks for willing to work (later) on it!

0x72D0:
I test the performance for projectionProfile, this is the performance when horizontal is true:

![screenshot_20181110_074156](https://user-images.githubusercontent.com/16087328/48301461-959ed880-e4bc-11e8-8791-6085fd99d0ab.png)

and when horizontal is false:

![screenshot_20181110_073934](https://user-images.githubusercontent.com/16087328/48301463-a2233100-e4bc-11e8-8a6c-33dfbe1022cd.png)

the function_pool performance seems to be better on larger image, but worst on small image.

ihhub:
This is what I expected. Multithreaded version is slower for small images because mutex synchronization takes some time.

ihhub:
@0x72D0 are you going to implement SSE and NEON as well?

0x72D0:
yeah sure, I think, I'm gonna do sse today, but for MEON it migth take a little bit of time before I completed it...

ihhub:
I was wrong regarding thread pool (function pool). It's slower because at the end of operation it copies all results into one buffer. This is done because you cannot pass a part of std::vector. We need to fix such limitation in future.

0x72D0:
but if I understand well, this limitation only target the ProjectionProfile and the Accumulate function, so it's not that bad :)

ihhub:
[...]

PR #248, opened 2018-11-08
Add AVX2.0 implementation of ProjectionProfile function

0x72D0:
PR for #116

the performance when horizontal is false:

screenshot_20181108_133057

the performance when horizontal is true:

screenshot_20181108_132833

all units tests pass.

I also fix a bug with the random value, when I was doing the units tests for projection profile, the randomization of the horizontal variable always gave me true.


0x72D0:
ok I revert it :)

test/performance_tests/performance_test_image_function.cpp
```diff
@@ -327,6 +335,7 @@ namespace image_function
     SET_FUNCTION( LookupTable        )
     SET_FUNCTION( Maximum            )
     SET_FUNCTION( Minimum            )
```
ihhub: Please a test for function_pool as well as it uses your new code internally. It should work even faster. Additionally, please align **)** symbol with other SET_FUNCTION lines by white space shift :)

0x72D0:
I don't know why but it's not faster with function_pool:

screenshot_20181108_224505

ihhub:
> I don't know why but it's not faster with function_pool:
> 
> screenshot_20181108_224505

Could you please verify that your new AVX code is called via function pool? It might be some missing hookups but logically adding one line for function table (which you did) fixes the problem.

0x72D0:
also I forgot to warn you about this code:

https://github.com/0x72D0/penguinV/blob/f01e1897a2ccef26d404933e4b52463923ac92c2/test/performance_tests/performance_test_image_function.cpp#L228

since I don't know how to test the true and false horizontal at the same time, I fixed it to false

ihhub:
@0x72D0 thanks a lot for your contribution to the library!

Issue #41, opened 2017-07-03
Support of libpng for Windows


ihhub:
Windows doesn't have third-party support of libpng (and zlib). We need to do a nice code which would be easy to use

Correlates with #7

Issue #156, opened 2018-10-19
Add unit tests for PNG file support like it's done for BITMAP


ihhub:
**PNG** support is made through `libpng` library usage. Currently we have a code which is supported only Unix OSes (`FileOperation/png_file.h`). We need to add unit tests similar to those for Bitmap (`test/unit_tests/unit_test_bitmap.cpp`).

PR #257, opened 2018-11-14
Adding libpng support

yannlabou:
Should work as it is on most UNIX platforms


ihhub:
Give me please some time to fix cpp file.

ihhub:
Let's fix installation on AppVeyor first. **libpng** requires **zlib** library. We need to install **zlib** and **libjpeg** before running CMake. You could follow [this link](https://stackoverflow.com/questions/36992243/appveyor-c-cmake-png-package-compilation-error) to fix this problem.

Second thing is to do not fail for normal CMake usage on user's machine. There is a way to do it: we could set an extra flag (macro) which is used in cpp file. If it's Windows system or this flag is not set then we don't compile a code which is inside `ifdef ... endf` block. We could just return an exception for each function.

yannlabou:
> Let's fix installation on AppVeyor first. **libpng** requires **zlib** library. We need to install **zlib** and **libjpeg** before running CMake. You could follow [this link](https://stackoverflow.com/questions/36992243/appveyor-c-cmake-png-package-compilation-error) to fix this problem.
> 
> Second thing is to do not fail for normal CMake usage on user's machine. There is a way to do it: we could set an extra flag (macro) which is used in cpp file. If it's Windows system or this flag is not set then we don't compile a code which is inside `ifdef ... endf` block. We could just return an exception for each function.

I'm on it.

ihhub:
I stopped x32 build for AppVeyor as x64 build failed.

test/unit_tests/CMakeLists.txt
```diff
@@ -28,6 +29,72 @@ set(SRCS
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+if(WIN32)
+    include(ExternalProject)
+    set(zlib_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/external/zlib_archive)
+    set(ZLIB_INSTALL ${CMAKE_CURRENT_BINARY_DIR}/zlib/install
```
ihhub: Should be **)** at the end?

yannlabou: I fixed it :)

yannlabou:
> I stopped x32 build for AppVeyor as x64 build failed.



> I stopped x32 build for AppVeyor as x64 build failed.

Is there a way to trigger only Appveyor? We know that it works on Travis.

ihhub:
> > I stopped x32 build for AppVeyor as x64 build failed.
> 
> Is there a way to trigger only Appveyor? We know that it works on Travis.

sadly, no. I'll stop Travis-CI build.

yannlabou:
> > > I stopped x32 build for AppVeyor as x64 build failed.
> > 
> > 
> > Is there a way to trigger only Appveyor? We know that it works on Travis.
> 
> sadly, no. I'll stop Travis-CI build.

OK

test/unit_tests/CMakeLists.txt
```diff
@@ -28,6 +29,72 @@ set(SRCS
...
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${zlib_INCLUDE_DIR}
+        DEPENDS zlib)
+    add_custom_target(zlib_copy_headers_to_destination 
+        DEPENDS zlib_create_destination_dir)
+    foreach(header_file ${ZLIB_HEADERS})
+        add_custom_command(TARGET zlib_copy_headers_to_destination PRE_BUILD
+            COMMAND ${CMAKE_COMMAND} -E copy_if_different ${header_file} ${zlib_INCLUDE_DIR})
+    endforeach()
+
+    set(png_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/external/png_archive)
```
ihhub: `Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR)`

Why is `PNG_PNG_INCLUDE_DIR`?

yannlabou: `find_package` is not even aware of `ExternalProject_Add`. No worries, I'm still working on it.

ihhub:
Now the problem is that `png.h` file is not in build paths...

yannlabou:
> Now the problem is that `png.h` file is not in build paths...

`target_include_directories` was apparently not successful. Let me think about it.

src/FileOperation/png_file.cpp
```diff
@@ -1,10 +1,5 @@
 #include <stdlib.h>
-
-#ifdef _WIN32
-
-#else
-    #include <png.h>
-#endif
+#include <png.h>
```
ihhub: Need to find how to set it for Windows.

yannlabou:
1st commit 12 hours ago! This one was hard. :sparkles:

ihhub:
@yannlabou Great job!

We would need to make 2 changes in future:
1. Move **libpng** download and installation code on Windows to a common location as it wiuld be used not only for unit tests.
2. Introduce a macro which should be set if we cannot install **libpng** on machine (for example, no internet). If this macro is set we would hide png code for we won't have compilation error. This part I can do within this pull request.

.travis.yml
```diff
@@ -12,7 +12,7 @@ addons:
 matrix:
   include:
     - os: linux
-      env: CUDA=8.0.61-1 CMAKE_OPTIONS="-DPENGUINV_BUILD_TEST=ON"
+      env: CUDA=8.0.61-1 CMAKE_OPTIONS="-DPENGUINV_BUILD_TEST=ON -DPENGUINV_ENABLE_PNG_SUPPORT=ON -DPENGUINV_USE_EXTERNAL_PNG=OFF"
```
ihhub: I thought we just need to create another file and move the logic there and then include this file in CMake file for unit tests?

yannlabou: I moved the logic in extra cmake files or maybe I'm misunderstanding you. Anyway I have issue on Appveyor I still need to fix.

ihhub: What I meant is: we move installation of libpng for Win32 and Unix in a separate file. If we are unable to install it we set a flag (macro), let's call it `PENGUINV_DISABLE_PNG`. This is C++ macro, not environment variable.

ihhub: Something like this:
Create a separate CMake file with initial logic which was in Cmake for unit tests. Put it, for example, in `script/cmake_lib` folder. In unit tests Cmake add a line of code `include("${CMAKE_CURRENT_LIST_DIR}/../../script/cmake_lib/png.cmake")`.

yannlabou: > Something like this:
> Create a separate CMake file with initial logic which was in Cmake for unit tests. Put it, for example, in `script/cmake_lib` folder. In unit tests Cmake add a line of code `include("${CMAKE_CURRENT_LIST_DIR}/../../script/cmake_lib/png.cmake")`.

I created two CMake modules `cmake/external_zlib.cmake` and `cmake/external_png.cmake` and I'm calling
```
include(external_zlib)
include(external_png)
```
in `test/unit_tests/CMakeLists.txt`. For it's exactly what you're requiring. I can rename the files/folders later on if the builds are successful.

ihhub: Do it even better: why don't you include `external_zlib` inside `external_png` in unit tests cmake we include only one as png requires zlib? :)

yannlabou: > The idea behind is that we enable `png` support by default but disable it when we cannot install the library :) It's an opposite idea what you have right now.
> It's better to do this way because as an user won't want to tune some variables and etc.

PNG support is enabled by default. We are not installing anything on UNIX platforms. On Windows, as long as CMake is able to download the library, everything is fine. I just don't know how to indicate a CMake download failure with a macro. Feel free to edit the pr, if you know how to do it. Currently, I'm not even sure if moving the CMake scripts to separate files will work on Windows.

ihhub:
@yannlabou I can make you an a user with an ability to stop or run builds on AppVeyor but I need your email to send an invitation. You could write it here and once I send an invitation I remove the message.

.travis.yml
yannlabou: > Do it even better: why don't you include `external_zlib` inside `external_png` in unit tests cmake we include only one as png requires zlib? :)

Will merge as soon as Appveyor builds properly.

yannlabou:
> @yannlabou I can make you an a user with an ability to stop or run builds on AppVeyor but I need your email to send an invitation. You could write it here and once I send an invitation I remove the message.

That would significantly speed up the work :). yann.labou<at>gmail.com

yannlabou:
It looks like I can't cancel the build.

ihhub:
Please try now.

yannlabou:
> Please try now.

👍

yannlabou:
Hi @ihhub. I think it's perfectly fine to review this 50 commits pr :).

ihhub:
Okay. Will do in a while :)

cmake/external_png.cmake
```diff
@@ -0,0 +1,131 @@
+include(ExternalProject)
```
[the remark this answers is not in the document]
yannlabou: I see your point and I will make the changes. My arguments was that I never considered cmake files as scripts :). It has also been established in many open source projects to put cmake modules in a folder called `cmake/modules/*.cmake` or `cmake/*.cmake`. It might be due to the fact that `CMake` itselfs ships all its modules into `<cmake_install_path>/share/cmake/Modules`.

test/unit_tests/CMakeLists.txt
```diff
@@ -58,9 +58,23 @@ else()
...
+    include(external_png)
+    target_link_libraries(unit_tests PRIVATE ZLIB_EXTERNAL PNG_EXTERNAL)
+else()
+    find_package(PNG REQUIRED)
+    target_link_libraries(unit_tests PRIVATE PNG::PNG)
+endif()
+
+if (WIN32)
+    # Quiet warnings related to fopen on MSVC
+    target_compile_definitions(unit_tests PRIVATE _CRT_SECURE_NO_WARNINGS)
```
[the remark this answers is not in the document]
yannlabou: Definitely as we are treating warnings as errors. `MSVC` provided years ago a non-portable version of all functions that use buffers to improve security :(.

test/unit_tests/CMakeLists.txt
```diff
@@ -58,9 +58,23 @@ else()
...
 
 add_executable(unit_tests ${SRCS})
-target_link_libraries(unit_tests
-                      PRIVATE unit_tests_options
-                              unit_tests_warnings
-                              Threads::Threads)
+target_link_libraries(unit_tests PRIVATE unit_tests_options unit_tests_warnings Threads::Threads)
+
+if(${PENGUINV_ENABLE_PNG_SUPPORT})
+    target_sources(unit_tests PRIVATE ${LIB_DIR}/FileOperation/png_file.cpp)
```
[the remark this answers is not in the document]
ihhub: I don't like them either. They are a root of big problems sometimes but like in this situation they are only suitable way.

test/unit_tests/CMakeLists.txt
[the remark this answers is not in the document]
ihhub: Okay, I'll check logs on AppVeyor and fix all warnings so them we can remove this.

cmake/external_png.cmake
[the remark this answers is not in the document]
ihhub: That's okay then. It's not a mandatory change after all :)

cmake/install_png.cmake
```diff
@@ -0,0 +1,135 @@
+include(ExternalProject)
+if(NOT ${PENGUINV_USE_EXTERNAL_PNG})
```
ihhub: My idea for this file is:
```bash
include(ExternalProject)

option(PENGUINV_ENABLE_PNG_SUPPORT "Enable support of libpng" ON)
option(PENGUINV_USE_EXTERNAL_PNG "Download libpng and build from source" OFF)

if(NOT ${PENGUINV_ENABLE_PNG_SUPPORT})
    find_package(PNG REQUIRED)
    if(NOT PNG_FOUND)
        set(PENGUINV_USE_EXTERNAL_PNG "ON")
    endif()
endif()

if(${PENGUINV_ENABLE_PNG_SUPPORT})
    current code for external support
```

Explanation:
1. We declare variable `PENGUINV_ENABLE_PNG_SUPPORT` only when we include this file, intentionally knowing this. Currently if I want to run cmake file only for unit tests, it would fail due to unset variables.
2. By default we try to find png package. If we failed to find then we would try to use external source of installation.
3. We remove variable declarations from main Cmake file as we declare them here.
4. We remove variable setup in `appveyor.yml` file as well.

Of course we might need to move
```bash
if(NOT CMAKE_MODULE_PATH)
    set(CMAKE_MODULE_PATH)
endif()
list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
```
logic in cmake for unit tests. This is I'm not sure about right now.

ihhub:
This pull request would fix #41 after the commit.

yannlabou:
> This pull request would fix #41 after the commit.

Proud to fix an one year old issue :).

ihhub:
After committing this pull request we could freely start to work on #156 as well.

ihhub:
Hi @yannlabou do you need any help in this? I see we're almost at the end :)

test/unit_tests/CMakeLists.txt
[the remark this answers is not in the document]
ihhub: Do you agree to set here the macro which I added in png_file.cpp file?

test/unit_tests/CMakeLists.txt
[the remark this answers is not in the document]
ihhub: I'll do it in a separate pull request.

test/unit_tests/CMakeLists.txt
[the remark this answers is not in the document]
ihhub: My point of view: soon we'll add unit tests for png. In this case we have to add file inclusion into unit_tests.cpp as well as call needed function for tests. Png tests itself must include png_file.h. If we don't do it we would have compilation error.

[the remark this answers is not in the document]
yannlabou: I believe
```
if(${PENGUINV_ENABLE_PNG_SUPPORT})
    target_sources(unit_tests PRIVATE ${LIB_DIR}/FileOperation/png_file.cpp)
endif()
```
is cleaner. The user shouldn't reason about the macro in the code later.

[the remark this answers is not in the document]
yannlabou: > My point of view: soon we'll add unit tests for png. In this case we have to add file inclusion into unit_tests.cpp as well as call needed function for tests. Png tests itself must include png_file.h. If we don't do it we would have compilation error.

Fair point. Let's do the changes. I think restructuring the library project layout in a future version will solve these kind of issues.

cmake/install_png.cmake
yannlabou: Done in 07d979f

ihhub:
@yannlabou great work! We did it !

Issue #247, opened 2018-11-07
Change return value of example projects from 0 to 1 and 2


ihhub:
All examples in `examples` directory have contain `main` function body which consist of `try ... catch` block:
```cpp
try // <---- do not forget to put your code into try.. catch block!
{
    // code
}
catch( const imageException & ex ) {
    // uh-oh, something went wrong!
    std::cout << "Exception " << ex.what() << " raised. Do your black magic to recover..." << std::endl;
    // your magic code must be here to recover from bad things
    return 0;
}
catch( ... ) {
    // uh-oh, something terrible happen!
    std::cout << "Something very terrible happen. Do your black magic to recover..." << std::endl;
    // your magic code must be here to recover from terrible things
    return 0;
}

std::cout << "Everything went fine." << std::endl;

return 0;
```

We need to modify every code into:
```cpp
try // <---- do not forget to put your code into try.. catch block!
{
    // code
}
catch( const std::exception & ex ) { // uh-oh, something went wrong!
    std::cout << "Exception " << ex.what() << " raised. Closing the application..." << std::endl;
    return 1;
}
catch( ... ) { // uh-oh, something terrible happen!
    std::cout << "Generic exception raised. Closing the application..." << std::endl;
    return 2;
}

std::cout << "Application ended correctly." << std::endl;
return 0;
```

Notable changes:
- return code are different
- we catch `std::exception` instead of `imageException `

kgorecki:
@ihhub what about some cool enumerator for error codes?

ihhub:
Each example has independant code. You're suggesting to duplicate the enumeration in each example. This isn't reasonable.

kgorecki:
I'd rather suggest something like this in some header file to be included in examples
enum class PenguinErrorCodes
{
  NoError = 0,
  GenericException,
  ImageException
};

ihhub:
We print a message before returning something so we could even just return 1 for both cases without enumeration.

ihhub:
We need to keep example code as sumple as possible so people won't turn their backs. Enumeration just complicates a little the understanding of the code if it would be in a separate file or somewhere else.

kgorecki:
Sure, it was just a suggestion. If nobody else wants to get this one, I can take it.

ihhub:
Okay, thanks!

ihhub:
@kgorecki please don't be upset if I wrote something wrong. I've just explain my point of view :)

kgorecki:
[...]

PR #260, opened 2018-11-16
Changed error codes for examples

kgorecki:
Resolves issue #247

### DIFF
--- a/examples/bitmap_operation/example_bitmap_operation.cpp
+++ b/examples/bitmap_operation/example_bitmap_operation.cpp
@@ -24,9 +24,8 @@ int main()
         // Second way to do
         method2();
     }
-    catch( imageException & ex ) { // uh-oh, something went wrong!
-        std::cout << "Exception " << ex.what() << " raised. Do your black magic to recover..." << std::endl;
-        // your magic code must be here to recover from bad things
+    catch( const std::exception & ex ) { // uh-oh, something went wrong!
+        std::cout << "Exception " << ex.what() << " raised. Closing the application..." << std::endl;
         return 1;
     }
     catch( ... ) { // uh-oh, something terrible happen!
@@ -34,8 +33,7 @@ int main()
         return 2;
     }
 
-    std::cout << "Everything went fine." << std::endl;
-
+    std::cout << "Application ended correctly." << std::endl;
     return 0;
 }
 
--- a/examples/bitmap_operation/example_bitmap_operation.cpp
+++ b/examples/bitmap_operation/example_bitmap_operation.cpp
@@ -24,17 +24,18 @@ int main()
         // Second way to do
         method2();
     }
-    catch( const imageException & ex ) {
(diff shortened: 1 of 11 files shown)


examples/bitmap_operation/example_bitmap_operation.cpp
```diff
@@ -24,17 +24,18 @@ int main()
         // Second way to do
         method2();
     }
-    catch( const imageException & ex ) {
-        // uh-oh, something went wrong!
+    catch( imageException & ex ) { // uh-oh, something went wrong!
```
ihhub: Please remove this catch block from each file. As it's mentioned in the issue we replace `imageException` by `std::exception`  in catch block. `imageException` class is inherited from `std::exception` base class so catching base class exception would catching inherited (class) exception too.

examples/bitmap_operation/example_bitmap_operation.cpp
```diff
@@ -24,17 +24,14 @@ int main()
...
```
ihhub: Please follow the structure which I wrote in the original issue:
```cpp
catch( const std::exception & ex ) { // uh-oh, something went wrong!
    std::cout << "Exception " << ex.what() << " raised. Closing the application..." << std::endl;
    return 1;
}
catch( ... ) { // uh-oh, something terrible happen!
    std::cout << "Generic exception raised. Closing the application..." << std::endl;
    return 2;
}

std::cout << "Application ended correctly." << std::endl;
return 0;
```

Just copy-paste the code in all files :)
Only for below files we don't need `std::cout << "Application ended correctly." << std::endl;`:
```bash
examples/cuda/pinned_memory/pinned_memory.cpp
examples/qt_framework/qimage_display/main.cpp
examples/raspicam/main.cpp
examples/thread_pool/example_thread_pool.cpp
```
Above files output or show something so we don't need to output last message which is an indicator that application completed it's work.

kgorecki: Sorry for that, I was in hurry with last commit and totally messed stuff.
I also removed unreachable "return 0;" from qt_framework/qimage_display/main.cpp

ihhub: No blame on you :) We all make mistakes ;)

kgorecki: True, but I should have followed the rule: first think, than do, than commit ;}

ihhub: I fixed that,

kgorecki:
@ihhub "return 0;" line in Qt example is unreachable now, so I'd suggest to take it away or change it to "return app.exec();" (current line 84).

ihhub:
@kgorecki yes, you're totally write. I revert it back.

ihhub:
@kgorecki thanks a lot for continuous support of th project!

Issue #261, opened 2018-11-17
Add cmake files for existing example projects


ihhub:
We have a bunch of example projects in `examples` directory:
```bash
bitmap_operation
blob_detection
cuda/image_function
cuda/pinned_memory
filters
function_pool
image_function
opencl/image_function
thread_pool
```

We need to add `cmake` files inside each project directory as well as add a simple description how to compile a project with `cmake` inside `README.md` file.

yannlabou:
/pull/271

Issue #166, opened 2018-10-21
Add CUDA check in CMake and add cmake files for unit and performance tests 


ihhub:
We do not have a check  for CUDA presence in a system via `CMake`. This is needed to add unit and performance tests into cmake files and make unified compilation of all tests.

jvce92:
Hi, if no one started working on it, I can take this issue

ihhub:
Sure.

yannlabou:
@jvce92 Do you need help on this?

jvce92:
Hi @yannlabou, I'm having some issues with linker errors that I'm trying to figure out. I've never used CMake with CUDA before and I struggled a bit using CMake native support.

Could you give me some guidance?

 Here is my CMakeLists.txt file https://github.com/jvce92/penguinV/blob/issue-166-add-cuda-check-and-cmake-files/test/performance_tests/cuda/CMakeLists.txt

ihhub:
@jvce92 could you please post message the errors? It'd be easier to find out what went wrong.

ihhub:
@jvce92 by the way, you don't need to compile `${LIB_DIR}/cuda/cuda_fft.cu` as it's not included in performance tests so you could remove `find_library(CUFFT_LIBRARY cufft ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})` line as well.

jvce92:
Pretty much every function in the test source files comes up as an undefined reference in the last linking step. Here is stderr output 

`CMakeFiles/performance_tests_cuda.dir/performance_test_helper_cuda.cu.o: In function `Performance_Test::Cuda_Helper::TimerContainerCuda::TimerContainerCuda()':
tmpxft_000019a0_00000000-4_performance_test_helper_cuda.cudafe1.cpp:(.text+0x71): undefined reference to `Performance_Test::BaseTimerContainer::BaseTimerContainer()'
tmpxft_000019a0_00000000-4_performance_test_helper_cuda.cudafe1.cpp:(.text+0xb0): undefined reference to `Performance_Test::BaseTimerContainer::~BaseTimerContainer()'
CMakeFiles/performance_tests_cuda.dir/performance_test_helper_cuda.cu.o: In function `Performance_Test::Cuda_Helper::TimerContainerCuda::~TimerContainerCuda()':
[...]

PR #271, opened 2018-11-18
Add CMakeLists.txt files for the examples

yannlabou:
Relates to #261


yannlabou:
Interesting... Build failed on [Travis Job](https://travis-ci.org/ihhub/penguinV/jobs/456757013) with
````
Scanning dependencies of target cuda_example_image_function
[ 66%] Building CXX object examples/cuda/image_function/CMakeFiles/cuda_example_image_function.dir/example_image_function.cpp.o
In file included from /home/travis/build/ihhub/penguinV/examples/cuda/image_function/example_image_function.cpp:6:0:
/home/travis/build/ihhub/penguinV/examples/cuda/image_function/../../../src/cuda/cuda_device.cuh:3:26: fatal error: cuda_runtime.h: No such file or directory
 #include <cuda_runtime.h>
                          ^
compilation terminated.
````
but report success.

File: examples/bitmap_operation/README.md [hunk 1]
```diff
@@ -1,15 +1,26 @@
-# How to compile
```
ihhub: We have to have 4 whitespaces at the end of needed lines in **md** file, otherwise file view is corrupted

ihhub:
@yannlabou can we compile an individual example after this changes? I mean like going to a specific directory of example and run cmake.

ihhub:
> @yannlabou can we compile an individual example after this changes? I mean like going to a specific directory of example and run cmake.

Why I'm asking it because users want to have an ability to compile any project within the library without depending on other Cmake files. It looks stupid from one side but from another sometimes user doesn't need to compile everything in order to run a single example project. The same idea applies for tests as well: we could compile only unit tests or only performance tests.

yannlabou:
> > @yannlabou can we compile an individual example after this changes? I mean like going to a specific directory of example and run cmake.
> 
> Why I'm asking it because users want to have an ability to compile any project within the library without depending on other Cmake files. It looks stupid from one side but from another sometimes user doesn't need to compile everything in order to run a single example project. The same idea applies for tests as well: we could compile only unit tests or only performance tests.

If you look carefully at the doc, I mentioned for example to build the `bitmap_operatien` example with 
```
cmake --build ./examples/bitmap_operation --config Release
```

What the user must definitely do is to configure the whole project first and therefore call `cmake ..` from the build folder of the root project directory. 
From a library perspective, it's necessary in order to avoid calling `find_package` everywhere and caring about `include_directories` and `link_libraries` mess. I agree with you and I'm fan of selective building  but there is always a trade-off between usability and library packaging. C++ is just not good at packaging. By the way we might consider [conan](https://conan.io/) the 'new' package manager. I talked to their devs at the conference and it's definitely worth the try. We spend too much time in the C++ community managing dependencies.

ihhub:
> If you look carefully at the doc, I mentioned for example to build the `bitmap_operatien` example with
> 
> ```
> cmake --build ./examples/bitmap_operation --config Release
> ```
> What the user must definitely do is to configure the whole project first and therefore call `cmake ..` from the build folder of the root project directory.
> From a library perspective, it's necessary in order to avoid calling `find_package` everywhere and caring about `include_directories` and `link_libraries` mess. I agree with you and I'm fan of selective building but there is always a trade-off between usability and library packaging. C++ is just not good at packaging. By the way we might consider [conan](https://conan.io/) the 'new' package manager. I talked to their devs at the conference and it's definitely worth the try. We spend too much time in the C++ community managing dependencies.

Ah, okay. Managing dependencies indeed is a problem. Introducing another way to do may turn users back from the library as it requires extra software to do it. We could consider this but we should always think from user perspective. For user an ideal software should run though one command or button :)

ihhub:
@yannlabou let's finish this pull request and concentrate on UI side as it's top important task for now ;)

examples/cuda/image_function/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(cuda_example_image_function
```
ihhub: For `.*cu` files we need to use a separate compiler **nvcc**. We have the same problem for CUDA unit tests: we use makefile for compilation. I recommend to don't spend time on CUDA right now and just do not add Cmake files for these 2 project examples.
We could address this problem in a separate issue.

CMakeLists.txt
```diff
@@ -7,6 +7,12 @@ if(NOT CMAKE_MODULE_PATH)
 endif()
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
+find_package(OpenCL)
+find_package(Qt5Widgets)
+find_package(CUDA)
```
ihhub: This change we could revert for now as we cannot compile CUDA

examples/CMakeLists.txt
```diff
@@ -1,4 +1,38 @@
...
+                                     -Wconversion
+                                     -Wsign-conversion
+                                     -O2
+                                     -march=native)
 endif()
+
+set(LIB_DIR ${CMAKE_SOURCE_DIR}/src)
+add_subdirectory(bitmap_operation)
+add_subdirectory(blob_detection)
+if(CUDA_FOUND)
```
ihhub: Let's remove this if else with cuda examples as of now or just comment these files

examples/function_pool/example_function_pool.cpp
```diff
@@ -135,7 +135,7 @@ void basic( const std::vector < PenguinV_Image::Image > & frame )
 void multithreaded( const std::vector < PenguinV_Image::Image > & frame )
 {
     // okay we setup 4 thread in global thread pool
-    Thread_Pool::ThreadPoolMonoid::instance().resize( 4 );
```
ihhub: That's why I propose to compile examples. Great catch!

ihhub:
Good morning, @yannlabou :)

We have an issue to addressing CUDA compilation #166 I'm proposing to fix it in that issue instead. What do you think?

ihhub:
I'll fix opencl_helper.cpp warning compilation.

yannlabou:
> @yannlabou let's finish this pull request and concentrate on UI side as it's top important task for now ;)

Sure. I thought, this would have been an easy one ;).

CMakeLists.txt
yannlabou: I agree

examples/CMakeLists.txt
yannlabou: OK

yannlabou:
> Good morning, @yannlabou :)
> 
> We have an issue to addressing CUDA compilation #166 I'm proposing to fix it in that issue instead. What do you think?

I fully agree. I actually forgot that one issue.

examples/cuda/image_function/CMakeLists.txt
yannlabou: That's indeed curious as I checked the `unit_tests/cuda/Makefile` and I wonder how this work because `THIRD_PARTY_DIR := $(LIB_DIR)/thirdparty/multicuda/src` point to a non-existent directory.

ihhub: I think I forgot to remove third party from makefile. The code was there but after a while I merged 2 projects

yannlabou:
@ihhub Do we need Qt on `Appveyor` or we build the Qt example only on `Ubuntu` and `MacOS`?

examples/cuda/image_function/CMakeLists.txt
ihhub: We don't need to commit this file as well as `examples/cuda/pinned_memory/CMakeLists.txt`. They are invalid for CUDA compilation so no need to commit them at all. Please remove them from the pull request.

ihhub:
Well done, @yannlabou ! Great work as usual!

PR #288, opened 2018-12-15
Renamed files in FileOperation and fixes makefile, CMake and VS project files

### LINKED ISSUE #239: Rename 4 files and fix makefile, CMake and VS project files after
We have 4 files in our library which we want to rename (for future library development):
```bash
src/FileOperation/bitmap.cpp
src/FileOperation/bitmap.h
src/FileOperation/png_file.cpp
src/FileOperation/png_file.h
```
new names and location for these files must be:
```bash
src/file/bmp_image.cpp
src/file/bmp_image.h
src/file/png_image.cpp
src/file/png_image.h
```
After that we need to fix file location in all needed makefiles, CMake files and VS project files which locate in `examples` and `src` directories (not so many as you could think :) ).

For a person willing to work on this issue: please do not hesitate to make a first pull request with what you think covers all needed files. If something would be missing we could always point it and correct.

aasimk2000:
Fixes #239


ihhub:
Hi, @aasimk2000 , thanks for making this pull request. I'll take a look into it soon.

ihhub:
@aasimk2000 thank you very much for your contribution to the project!

Issue #267, opened 2018-11-18
Add CUDA implementation of Rotate function


ihhub:
We have Rotate function implementation which locates in `src/image_function.h` and `src/image_function.cpp` files:
```cpp
void Rotate( const Image & in, double centerXIn, double centerYIn, Image & out, double centerXOut, double centerYOut, double angle );
```
We want to add CUDA implementation of the same function. Please put this implementation into `src/cuda/image_function_cuda.cuh` and `src/cuda/image_function_cuda.cu`. Use an existing code style of format to do it so it should pretty straightforward.

MatthewMcGonagle:
I'm interested in taking a look at this, but I have a question about something inside `Image_Function::Rotate()`. It concerns the following code:
``` cpp
                        // we use bilinear approximation to find pixel intensity value
                        double coeffX = posX - x;
                        double coeffY = posY - y;

                        double sum = (*inX) * (1 - coeffX) * (1 - coeffY) +
                            (*inX + 1) * (coeffX) * (1 - coeffY) +
                            (*inX + rowSizeIn) * (1 - coeffX) * (coeffY)+
                            (*inX + rowSizeIn + 1) * (coeffX) * (coeffY)+0.5;

                        (*outX) = static_cast<uint8_t>(sum);
```

Its seems to me that `sum` is supposed to be an average of four pixels (although I'm not quite sure I understand the 0.5 offset). However, I think there is an error in the precedence of stuff like `*inX + 1`. My understanding is that the intention is that this is `*(inX + 1)`, but this is actually `(*inX) + 1`.

For example, the following minimal example
``` cpp
#include<iostream>

int main() {

    int nums[3] = {1, 10, 100};
    int *node = nums;

    std::cout << "*node = " << *node << std::endl
              << "*node + 1 = " << *node + 1 << std::endl
              << "*(node + 1) = " << *(node + 1) << std::endl;
    return 0;
}
```

gives as output of
```
*node = 1
*node + 1 = 2
*(node + 1) = 10
```

MatthewMcGonagle:
Another question. 

It seems to me that quantities outside of the `for` loops , such as `cosAngle` or `inXPos`, should be calculated by `Image_Function_Cuda::Rotate()` before calling a kernel on the GPU. The computed quantities can just be passed into the kernel. So the threads of the kernel simply handle doing the `for` loops in parallel.

Does this seem good to you?

ihhub:
[...]

PR #290, opened 2018-12-18
Add Rotate() function into Cuda namespace

MatthewMcGonagle:
Added Image_Function_Cuda::Rotate() and __global__ rotateCuda() Cuda Kernel.

Related to issue #267 Add CUDA implementation of Rotate function.

### DIFF
--- a/src/cuda/image_function_cuda.cu
+++ b/src/cuda/image_function_cuda.cu
@@ -239,14 +239,15 @@ namespace
         }
     }
 
-    __global__ void rotateCuda( const uint8_t * in, uint32_t rowSizeIn, uint32_t inHeight, float inXStart, float inYStart,
-                                uint8_t * out, uint32_t rowSizeOut, uint32_t outHeight, float cosAngle, float sinAngle )
+    __global__ void rotateCuda( const uint8_t * in, uint32_t rowSizeIn, uint8_t * out, uint32_t rowSizeOut,
+                                float inXStart, float inYStart, uint32_t width, uint32_t height, 
+                                float cosAngle, float sinAngle )
     {
         uint32_t outX = blockDim.x * blockIdx.x + threadIdx.x;
         uint32_t outY = blockDim.y * blockIdx.y + threadIdx.y;
 
         // Only do something if this thread is for a valid pixel in the output
-        if ( outX < rowSizeOut && outY < outHeight ) {
+        if ( outX < width && outY < height ) {
             // Both input coordinates are shifted using the cosAngle, sinAngle, outX, and outY. The shift
             // comes from inverse rotating the horizontal and vertical iterations over the output.
 
@@ -263,7 +264,7 @@ namespace
             out = out + outY * rowSizeOut + outX;
 
             // Note that we will be taking an average with next pixels, so next pixels need to be in the image too
-            if ( inX < 0 || inX >= rowSizeIn - 1 || inY < 0 || inY >= inHeight - 1 ) {
+            if ( inX < 0 || inX >= width - 1 || inY < 0 || inY >= height - 1 ) {
                 *out = 0; // We do not actually know what is beyond the image, so set value to 0
             }
             else {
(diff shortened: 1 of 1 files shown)


src/cuda/image_function_cuda.cu
```diff
@@ -239,6 +239,59 @@ namespace
         }
     }
 
+    __global__ void rotateCuda( const uint8_t * in, uint32_t rowSizeIn, uint32_t inHeight, float inXStart, float inYStart, 
+                                uint8_t * out, uint32_t rowSizeOut, uint32_t outHeight, 
+                                float cosAngle, float sinAngle)
+    {
+        uint32_t outX = blockDim.x * blockIdx.x + threadIdx.x,
+                 outY = blockDim.y * blockIdx.y + threadIdx.y;
```
ihhub: Please use an explicit declaration of variable type in all cases as people might read that this is a single variable.

```cpp
const uint32_t outX = blockDim.x * blockIdx.x + threadIdx.x,
const uint32_t outY = blockDim.y * blockIdx.y + threadIdx.y;
```

Please change the same for below cases as well.

src/cuda/image_function_cuda.cu
```diff
@@ -239,6 +239,59 @@ namespace
         }
     }
 
+    __global__ void rotateCuda( const uint8_t * in, uint32_t rowSizeIn, uint32_t inHeight, float inXStart, float inYStart, 
+                                uint8_t * out, uint32_t rowSizeOut, uint32_t outHeight, 
+                                float cosAngle, float sinAngle)
```
ihhub: Move this code to previous line. We have a recommended line size as 150 characters so it should be fine.

src/cuda/image_function_cuda.cu
```diff
@@ -239,6 +239,59 @@ namespace
...
+    {
+        uint32_t outX = blockDim.x * blockIdx.x + threadIdx.x,
+                 outY = blockDim.y * blockIdx.y + threadIdx.y;
+       
+        // If this thread isn't for a valid pixel in the output image, then do nothing. 
+
+        if ( outX >= rowSizeOut || outY >= outHeight )
```
ihhub: Please change this condition to opposite and put the code inside the condition. `return` is not recommended to use in GPU kernels as all threads expect to finish their work at the end of kernel's body. So the code would be like this:
```cpp
if ( outX < rowSizeOut && outY < outHeight ) {
    // your code below
}
```

src/cuda/image_function_cuda.cu
```diff
@@ -239,6 +239,59 @@ namespace
...
+        if ( outX >= rowSizeOut || outY >= outHeight ) 
+            return;
+
+        // Both input coordinates are shifted using the cosAngle, sinAngle, outX, and outY. The shift
+        // comes from inverse rotating the horizontal and vertical iterations over the output.
+
+        // Note that inverse rotation of horizontal motion is ( cos(angle), -sin(angle)),
+        // and the inverse rotation of vertical motion is ( sin(angle), cos(angle) ).
+
+        float exactInX = inXStart + cosAngle * outX + sinAngle * outY,
```
ihhub: Please use `const` word wherever it is possible in variable initialisation :)

src/cuda/image_function_cuda.cu
MatthewMcGonagle: Okay, done.

src/cuda/image_function_cuda.cu
MatthewMcGonagle: I moved `float cosAngle, float sinAngle)` to line above.

src/cuda/image_function_cuda.cu
MatthewMcGonagle: I used `const` where appropriate. I also removed the pointer variables `inMem` and `outMem` from `rotateCuda()` as they are unnecessary. We can just use `in` and `out` directly.

src/cuda/image_function_cuda.cu
MatthewMcGonagle: Okay, done.

PR #291, opened 2018-12-21
Projection profile neon

0x72D0:
PR for #116 
here's the performance on my raspberry pi when horizontal is false:
screenshot_20181221_124517

when horizontal is true:
screenshot_20181221_130029


0x72D0:
I will try to fix the performance when horizontal is true

0x72D0:
it seem that the vaddw intrinsics have some performance issues, now here's the new performance:
screenshot_20181221_135657

ihhub:
@0x72D0 thank you very much for continuous support of the project!

PR #303, opened 2018-12-23
Issue 277 line representation 2d

yannlabou:


src/math/math_base.h
```diff
@@ -105,5 +106,70 @@ struct PointBase3D : public PointBase2D<_Type>
     _Type z;
 };
 
+template <typename _Type>
+struct LineBase2D
```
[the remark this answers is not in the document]
yannlabou: The points are not meant as start and end points. Even with an angle, we still need at least two points to represent a line in a 2D space. Could you please provide a drawing to illustrate your point?

src/math/math_base.h
```diff
@@ -105,5 +106,70 @@ struct PointBase3D : public PointBase2D<_Type>
...
+{
+    LineBase2D(const PointBase2D<_Type> & _p1, const PointBase2D<_Type> & _p2)
+        : p1( _p1 ), p2( _p2 )
+    { }
+
+    LineBase2D(const PointBase2D<_Type> & _p, double angleRadians)
+    {
+        this->p1 = PointBase2D<_Type>(_p.x, _p.y);
+        this->p2 = PointBase2D<_Type>(std::cos(angleRadians) * length(), std::sin(angleRadians) * length());
```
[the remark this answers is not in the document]
yannlabou: A very good point. This is indeed unfortunate to do this that way.

yannlabou:
> I am suggesting to make this structure as a class with private point and angle members. We would have a function to set angle (which should be in a range of [-pi; pi] with needed normalisation), a function to set point. There shouldn't be function length as line has an infinite length.

I would make the changes after we agreed on the line representation. See my previous comment.
The function `length()` was rather meant here as helper function and not the "length" of a line.
> Function distance would return a distance from a point to line. Function intersect could be overloaded: one accepts only line while another - line and point as a reference to be set.

With `distance`, do you mean the shortest one between a point and the current line?
> Operator == should check angle and distance from one line point to another line. Operator += shouldn't exist.
> 

operator+= was meant as translation but I could remove it if not wanted

src/math/math_base.h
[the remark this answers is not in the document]
ihhub: line

Let's assume that point P0 is fixed so I can move point P1 along the line to any place and the line as well as angle would be the same.
I think to solve such problem we might introduce a class Vector which represents geometric vector from which we could easily find an angle. How about this solution? By the way vector and norm to vector would help to find a shortest distance to line and plane (in future).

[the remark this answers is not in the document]
ihhub: Let me write a skeleton of vector... few minutes please...

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        {
+            *point = p1 + a * na;
+            return true;
+        }
+
+        return false;
+    }
+
+    PointBase2D<_Type> p1;
+    PointBase2D<_Type> p2;
```
ihhub: Let's assume in this case that this kind of vector :) This is not a bug, just expressing my thoughts

src/math/math_base.h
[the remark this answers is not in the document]
ihhub:
```cpp
p2 = (p1.x + cos(angle), p1.x + sin(angle));
```

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+
+    LineBase2D(const PointBase2D<_Type> & _p, double angleRadians)
+    {
+        this->p1 = PointBase2D<_Type>(_p.x, _p.y);
+        this->p2 = PointBase2D<_Type>(std::cos(angleRadians) * length(), std::sin(angleRadians) * length());
+    }
+
+    bool operator == ( const LineBase2D & line ) const
+    {
+        return p1 == line.p1 && p2 == line.p2;
```
ihhub: This function is invalid as two lines might have different points but all 4 points in result sit on one line.

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+    }
+
+    LineBase2D & operator += ( const PointBase2D<_Type> & point )
+    {
+        return LineBase2D(this->p1 + point, this->p2 + point);
```
ihhub: I assume it is a shift operation for line, isn't it?

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+    }
+
+    double length() const
```
ihhub: I'm not sure that we need this function as no real purpose for it.

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+    {
+        const double x = static_cast<double>(p2.x) - static_cast<double>(p1.x);
+        const double y = static_cast<double>(p2.y) - static_cast<double>(p1.y);
+        return std::sqrt(x * x + y * y);
+    }
+
+    double angle() const
+    {
+        if (p1 == p2) throw std::invalid_argument("Points p1 and p2 must be different!");
```
ihhub: Let's add a function which validates line:
```cpp
bool valid() const
{
    return (p1 != p2);
}
```

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        const double x = static_cast<double>(p2.x) - static_cast<double>(p1.x);
+        const double y = static_cast<double>(p2.y) - static_cast<double>(p1.y);
+        const double angleRadians = std::atan2(y, x);
+        return angleRadians < 0 ? angleRadians + 2 * pvmath::pi : angleRadians;
+    }
+
+    bool intersect(const LineBase2D & line, PointBase2D<_Type> * point) const
```
ihhub: We could have 2 functions as I mentioned earlier: one with 1 parameter and another one with 2 parameters so we don't need to use pointer in this case.

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        // based on Graphics Gems III, Faster Line Segment Intersection, p. 199-202
+        // http://www.realtimerendering.com/resources/GraphicsGems/gems.html#gemsiii
+        const PointBase2D<_Type> a = p2 - p1;
+        const PointBase2D<_Type> b = line.p1 - line.p2;
+        const PointBase2D<_Type> c = p1 - line.p1;
+
+        const double denominator = a.y * b.x - a.x * b.y;
+        if (pvmath::isEqual<_Type>(denominator, 0))
+        {
+            return false; // parallel
```
ihhub: Shall we add a function to check whether lines are parallel?

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        }
+
+        const double inverse_denominator = 1.0 / denominator;
```
ihhub: Please rename into **inverseDenominator**.

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        const double na = (b.y * c.x - b.x * c.y) * inverse_denominator;
```
ihhub: Do we need to declare inverse denominator if it is used only one time?

ihhub:
> With `distance`, do you mean the shortest one between a point and the current line?

Yes.

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        this->p1 = PointBase2D<_Type>(_p.x, _p.y);
+        this->p2 = PointBase2D<_Type>(std::cos(angleRadians) * length(), std::sin(angleRadians) * length());
+    }
+
+    bool operator == ( const LineBase2D & line ) const
+    {
+        return p1 == line.p1 && p2 == line.p2;
+    }
+
+    LineBase2D & operator += ( const PointBase2D<_Type> & point )
```
ihhub: Please just add a comment here stating that this is `translation` function.

src/math/math_base.h
```diff
@@ -105,5 +106,73 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        return std::sqrt(x * x + y * y);
+    }
+
+    double angle() const
+    {
+        if (p1 == p2) throw std::invalid_argument("Points p1 and p2 must be different!");
+        const double x = static_cast<double>(p2.x) - static_cast<double>(p1.x);
+        const double y = static_cast<double>(p2.y) - static_cast<double>(p1.y);
+        const double angleRadians = std::atan2(y, x);
+        return angleRadians < 0 ? angleRadians + 2 * pvmath::pi : angleRadians;
```
ihhub: No need to convert to [0;2*pi] as this angle would be used somewhere in a code calling standard math functions.

ihhub:
I think there is no misunderstanding from anyone :) I think it's clear what we need to change before merge.
My apology that I pointed about angle, as a result your idea is still the same and even better :) Well, a constructive conversation always leads to a solution.

src/math/math_base.h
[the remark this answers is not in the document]
ihhub: I fixed it :)

src/math/math_base.h
```diff
@@ -105,5 +106,65 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        : p1( _p ), p2( _p.x + std::cos(angleRadians), _p.y + std::sin(angleRadians) )
+    { }
+
+    bool operator == ( const LineBase2D & line ) const
+    {
+        return p1 == line.p1 && p2 == line.p2;
+    }
+
+    // This is translation (shift) function
+    LineBase2D & operator += ( const PointBase2D<_Type> & point )
```
ihhub: This code is wrong. We create a temporary object inside function and return a reference to it after exiting the function.

If we want to implement translation function we need to use operator **+** and return a reference to `*this`.

ihhub:
I changed some obvious places to save some time. Waiting for your replies :)

src/math/math_base.h
```diff
@@ -105,5 +106,64 @@ struct PointBase3D : public PointBase2D<_Type>
...
 };
 
+template <typename _Type>
+struct LineBase2D
+{
+    LineBase2D( const PointBase2D<_Type> & _p1, const PointBase2D<_Type> & _p2 )
+        : p1( _p1 ), p2( _p2 )
+    { }
+
+    LineBase2D( const PointBase2D<_Type> & _p, double angleRadians )
```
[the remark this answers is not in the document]
yannlabou: I thought about it but we then also allow integral types as `angleRadians`. Do we want it?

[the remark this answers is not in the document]
ihhub: We couldn't cover situations but we have a good solution for this as concrete implementation of this function for specific type.

Logically only float and double types are used to represent a line while integral (like int, char and etc) are not really useful. At least I can't find any case for this...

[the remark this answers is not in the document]
ihhub: To explain further: not always necessary to use `double` just because it gives higher accuracy, sometimes `float` is enough. At least setting a type `float` for the angle in `float` implementation is reasonable.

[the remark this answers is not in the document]
yannlabou: So we use _Type?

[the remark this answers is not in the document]
ihhub: Yes, I'm suggesting this.

src/math/math_base.h
```diff
@@ -105,5 +106,65 @@ struct PointBase3D : public PointBase2D<_Type>
     _Type z;
 };
 
+template <typename _Type>
+struct LineBase2D
+{
+    LineBase2D( const PointBase2D<_Type> & _p1, const PointBase2D<_Type> & _p2 )
```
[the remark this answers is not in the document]
yannlabou: It won't for good reasons even not compile. MSVC19 accepts it though. https://godbolt.org/z/LuR44s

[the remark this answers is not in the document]
ihhub: Ah, okay. Wrong suggestion.

src/math/math_base.h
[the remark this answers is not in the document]
yannlabou: ok :)

src/math/math_base.h
yannlabou: > Let's add a function which validates line:
> 
> ```c++
> bool valid() const
> {
>     return (p1 != p2);
> }
> ```

Not needed anymore as `atan2` now handles the zero case.

src/math/math_base.h
yannlabou: Even with four points having the same orientation, if we represent two lines traversing each pair of points, they are still geometrically equal. Not sure how you want me to implement this but feel free to change it.

ihhub: If we have implemented `distance` function then comparing angles of both lines and measuring the distance from one point of line to another line will give us the correct result.

src/math/math_base.h
```diff
@@ -105,5 +106,61 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        {
+            return false; // parallel
+        }
+
+        const double na = (b.y * c.x - b.x * c.y) / denominator;
+        point = p1 + a * na;
+        return true;
+    }
+
+    PointBase2D<_Type> p1{};
```
ihhub: Why  do we need `{}` here? Isn't default constructor insufficient?

src/math/math_base.h
```diff
@@ -105,5 +106,61 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        *this = *this + point;
+        return *this;
+    }
+
+    double angle() const
+    {
+        return std::atan2(p2.y - p1.y, p2.x - p1.x);
+    }
+
+    bool intersect(const LineBase2D & line, const PointBase2D<_Type> & point) const
```
ihhub: `point` should be non-const reference as we modify it inside function.

src/math/math_base.h
```diff
@@ -105,5 +106,61 @@ struct PointBase3D : public PointBase2D<_Type>
     _Type z;
 };
 
+template <typename _Type>
+struct LineBase2D
+{
+    LineBase2D( const PointBase2D<_Type> & _p1, const PointBase2D<_Type> & _p2 )
+        : p1( _p1 ), p2( _p2 )
```
ihhub: Please use one variable initialisation per line like it was previously.

src/math/math_base.h
```diff
@@ -105,5 +106,61 @@ struct PointBase3D : public PointBase2D<_Type>
...
+    { }
+
+    bool operator == ( const LineBase2D & line ) const
+    {
+        return p1 == line.p1 && p2 == line.p2;
+    }
+
+    LineBase2D operator + ( const PointBase2D<_Type> & point )
+    {
+        return LineBase2D(this->p1 + point, this->p2 + point);
```
ihhub: Here is no need in explicit usage of `this`.

src/math/math_base.h
```diff
@@ -105,5 +106,67 @@ struct PointBase3D : public PointBase2D<_Type>
...
+
+    LineBase2D operator + ( const PointBase2D<_Type> & point )
+    {
+        return LineBase2D(p1 + point, p2 + point);
+    }
+
+    // This is translation (shift) function
+    LineBase2D & operator += ( const PointBase2D<_Type> & point )
+    {
+        *this = *this + point;
```
ihhub: Here is no mistake in code. Just thinking from performance side: such code requires 2 copy operations. If we simply do:
```cpp
p1 += point;
p2 += point;
return *this;
```
we need only 1 copy operation.

src/math/math_base.h
```diff
@@ -105,5 +105,81 @@ struct PointBase3D : public PointBase2D<_Type>
...
+    { }
+
+    LineBase2D( const PointBase2D<_Type> & _p, _Type angleRadians )
+        : p1( _p )
+        , p2( _p.x + std::cos(angleRadians), _p.y + std::sin(angleRadians) )
+    { }
+
+    bool operator == ( const LineBase2D & line ) const
+    {
+        return pvmath::isEqual<_Type>(angle(), line.angle()) && pvmath::isEqual<_Type>(distance(line.p1), 0);
```
ihhub: Actually, angle approach won't work for all cases. Let's assume we have a line with points A and B (for p1 and p2) and line with points B and A. Angles would be opposite. To solve this problem we need to have `parallel` function:
```cpp
return parallel( line ) && pvmath::isEqual<_Type>(distance(line.p1), 0);
```
then it's definitely the same line.

ihhub:
Hi, @yannlabou , I think everything looks very good. Just two things are remained:
1. Naming of `intersect` and `parallel` functions not very good but we can leave it as of now and rename later if we found better names :)
2. We need base unit tests for this class.

yannlabou:
> 2. We need base unit tests for this class.

Yeah. Proper unit tests are missing. I won't have time today and tomorrow for that.

src/math/math_base.h
```diff
@@ -105,5 +105,95 @@ struct PointBase3D : public PointBase2D<_Type>
...
+        // A distance from a point to a line can be calculated as:
+        // |A * x0 + B * y0 + C| / sqrt(A * A + B * B)
+        if ( p1 == p2 ) // there is no line
+            return 0;
+        const _Type A = (p2.y - p1.y);
+        const _Type B = -(p2.x - p1.x);
+        const _Type C = (p2.x * p1.y - p1.x * p2.y);
+        
+        const _Type numerator = (A * point.x + B * point.y + C);
+        return (numerator < 0 ? -numerator : numerator) / sqrt( A * A + B * B );
```
ihhub: I personally faced a situation in the past when I had to measure a distance from many points to a line. In such situation **sqrt** makes a significant impact into performance. My suggestion is to make p1 and p2 as a private members of the class. More precisely call them as **position** and **direction**. Position remains as it is - p1, while direction is normalised difference between p2 and p1. Now what we would have: sqrt equation would be equal to 1 because it's normalized. Coefficients A and B in distance function are already calculated. Variables A and B in parallel and intersect functions are calculated as well, they are equal to direction. Profit! :)

ihhub: If you agree with my suggestion I can make changes myself to save your time

yannlabou:
@ihhub you never sleep? It’s incredible how fast you’re. :)

ihhub:
@yannlabou thank you very much for working on this issue! Together we found the best solution for it ;)

PR #309, opened 2018-12-26
Installing explicitly Qt5.9 on Mac OS X

yannlabou:


yannlabou:
Won't do. Homebrew is designed to installed the latest version of a formula (package). Currently Qt 5.12.

Issue #274, opened 2018-11-22
CMake must fail for example compilation failure


ihhub:
We compile unit and performance tests as well as examples on Travis-CI and AppVeyor using CMake. Unfortunately, a compilation failure of any of example doesn't fail overall CMake compilation so we could be unaware of a problem. [This](https://travis-ci.org/ihhub/penguinV/jobs/458336343) is an example of such problem. Check for qt compiled example in the log (line **1265**).

ihhub:
@yannlabou any ideas how to fix it?

yannlabou:
I'll find out! :wrench:

PR #316, opened 2018-12-28
Travis-CI must fail for examples build

yannlabou:
relates to #274

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,11 @@ matrix:
       env: QT_BASE=latest CMAKE_OPTIONS=""
       before_install: source script/travis/install_qt5.sh
 
+before_script:
+  # Let CMake generate build files
+  - mkdir -p build && cd build
+  - cmake .. ${CMAKE_OPTIONS}
+
 script:
   - |
     if [[ "$STATIC_ANALYSIS" == "ON" ]]; then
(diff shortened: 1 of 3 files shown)


ihhub:
So the logic behind: we build CMake in `before script` which logically fails a job while doing the same in `script` section doesn't fail it. Is it correct?

yannlabou:
> So the logic behind: we build CMake in `before script` which logically fails a job while doing the same in `script` section doesn't fail it. Is it correct?

Not really. I tried to reorganized the build steps a little bit since [cmake .. ${CMAKE_OPTIONS}](https://github.com/ihhub/penguinV/blob/c0b5984c3b50ecea8f64752884564c75f4c67d32/.travis.yml#L40) is only generating the Makefile(s).

The decisive change is actually [set -e/set +e](https://github.com/ihhub/penguinV/blob/c0b5984c3b50ecea8f64752884564c75f4c67d32/.travis.yml#L53). `cmake --build .` wraps calls to `make` or `msbuild` and I had the feeling Travis had issues reading the return code on failure. `set -e`exits immediately on failure as I can see [here](https://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin).

ihhub:
Nice then! Well done job!

Issue #207, opened 2018-10-29
SIMD implementation of Hough Transform


ihhub:
We have _Hough Transform_ implementation of CPU (`src/math/hough_transform.h`). We want to add SIMD (**NEON**, **AVX** or **SSE**) implementation of the same function. The finest part for SIMD integration is this loop:
```cpp
for ( ; point != pointEnd; ++point, ++distanceVal )
    (*distanceVal) = point->x * sinVal + point->y * cosVal;
```

1. Create new files named `src/math/hough_transform_simd.h` and `src/math/hough_transform_simd.cpp`
2. Put the function under `Image_Function_Simd` namespace
3. Don't forget SIMD detection code which you could find in `src/image_function_simd.h` and `src/image_function_simd.cpp`

ihhub:
Hi, @0x72D0 would you like to work on it?

0x72D0:
yeah sure

0x72D0:
@ihhub I have some problem with c++ namespace, I try to use the simd::actualSimdType() in my hough_transform_simd.cpp but for some reason g++ say that actualSimdType is not a member of 'simd' here's the block of line:
```namespace Image_Function_Simd
{
    bool HoughTransform( const std::vector< Point2d > & input, double initialAngle, double angleTolerance, double angleStep, double lineTolerance,
                         std::vector< Point2d > & outOnLine, std::vector< Point2d > & outOffLine )
    {
        return simd::HoughTransform( input, initialAngle, angleTolerance, angleStep, lineTolerance, outOnLine, outOffLine, simd::actualSimdType() );
    }
}
```
do you know the problem? you can see the full code here: https://github.com/0x72D0/penguinV/tree/hough_transform_AVX

ihhub:
`simd::actualSimdType()` is a function which locates in `image_function_simd.cpp` file (lines 1590-1608) meaning that it is unaccessible. You have to define your own in `hough_transform_simd.cpp` file or move the code to some common place/file. I'd recommend to use second way but we should handle it gently.

ihhub:
@0x72D0 I'll make some changes in code to enable a proper support of SIMD across all files. I'll let you know once it's finished.

ihhub:
This issue relates to #216 .

ihhub:
@0x72D0 now you can complete this task with ease :)

ihhub:
Hi, @0x72D0 are you going to complete this task or have a try with Shift function?

[...]

PR #322, opened 2018-12-31
Avx implementation of Hough transform

0x72D0:
PR for #207 

with performance tests for the double function only:
screenshot_20181231_172006


src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        FindDistance( input, distance, cosVal, sinVal, inputPointCount );
+    }
+
+    template <>
+    void FindDistanceSimd< float >( const std::vector< PointBase2D< float > > & input, std::vector < float > & distance, float cosVal, float sinVal, const size_t inputPointCount, simd::SIMDType simdType )
+    {
+        if ( simdType == simd::avx_function ) {
+            #ifdef PENGUINV_AVX_SET
+            const uint32_t simdWidth = (inputPointCount*2) / (avx_float*2);
```
ihhub: We can get a size of input array directly without passing an extra variable.

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,276 @@
...
+            for(;point != PointEndSimd; point += avx_float, distanceVal += avx_float)
+            {
+                __m256 src1 = _mm256_loadu_ps(point);
+                point += avx_float;
+                __m256 src2 = _mm256_loadu_ps(point);
+
+                src1 = _mm256_mul_ps (src1, coeff);
+                src2 = _mm256_mul_ps (src2, coeff);
+
+                __m256 resul = _mm256_hadd_ps (src1, src2);
```
ihhub: `result` instead of `resul`

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,276 @@
...
+    template <>
+    void FindDistanceSimd< float >( const std::vector< PointBase2D< float > > & input, std::vector < float > & distance, float cosVal, float sinVal, const size_t inputPointCount, simd::SIMDType simdType )
+    {
+        if ( simdType == simd::avx_function ) {
+            #ifdef PENGUINV_AVX_SET
+            const uint32_t simdWidth = (inputPointCount*2) / (avx_float*2);
+            const uint32_t totalSimdWidth = simdWidth * (avx_float*2);
+            const uint32_t nonSimdWidth = (inputPointCount*2) - totalSimdWidth;
+
+            const float * point = reinterpret_cast<const float*>(input.data());
```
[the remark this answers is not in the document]
0x72D0: we need casting because input.data() return a PointBase2D<float> *, so we need to cast it to a const float*

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+        // sort points
+        const _Type cosVal = cos( angleVal );
+        const _Type sinVal = sin( angleVal );
+
+        _Type * distanceVal = distanceToLine.data();
+        const PointBase2D<_Type> * point = input.data();
+        const PointBase2D<_Type> * pointEnd = point + inputPointCount;
+
+        for ( ; point != pointEnd; ++point, ++distanceVal ) {
```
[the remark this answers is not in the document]
0x72D0: yes I forgot that

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,276 @@
...
+    template <typename _Type>
+    void FindDistanceSimd( const std::vector< PointBase2D< _Type > > & input, std::vector < _Type > & distance, _Type cosVal, _Type sinVal, const size_t inputPointCount, simd::SIMDType simdType )
+    {
+        FindDistance( input, distance, cosVal, sinVal, inputPointCount );
+    }
+
+    template <>
+    void FindDistanceSimd< float >( const std::vector< PointBase2D< float > > & input, std::vector < float > & distance, float cosVal, float sinVal, const size_t inputPointCount, simd::SIMDType simdType )
+    {
+        bool std_layout_check = std::is_standard_layout<PointBase2D<float>>::value;
```
ihhub: We can verify this one time by assigning to const variable outside this function.

0x72D0:
I did another test and it seems that the cpu version of the code is faster than the AVX one:
screenshot_20190101_195631

ihhub:
@0x72D0 let's see into code. Something could be wrong for example. Are you sure that AVX version reaches simd instructions?

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,271 @@
...
+    template <>
+    void FindDistanceSimd< float >( const std::vector< PointBase2D< float > > & input, std::vector < float > & distance, float cosVal, float sinVal, const size_t inputPointCount )
+    {
+        
+        simd::SIMDType simdType = simd::actualSimdType();
+
+        if ( simdType == simd::avx_function && float_layout_check ) {
+            #ifdef PENGUINV_AVX_SET
+            const uint32_t simdWidth = (inputPointCount*2) / (avx_float*2);
+            const uint32_t totalSimdWidth = simdWidth * (avx_float*2);
```
ihhub: What is we set `avx_float`value in 2 times bigger initially? We know that we handle float[2] array always.

0x72D0: no we also use `avx_float` in the for loop

ihhub:
@0x72D0 I made some changes in code. Let's see numbers you get for tests after it.

0x72D0:
no, AVX is still slower:
screenshot_20190101_205040

I'm sure it's the permutation instruction that slow down this algorithm, i'm gonna check if there's a replacement

ihhub:
@0x72D0 notice sigma value for AVX and for CPU. Anyway, AVX shouldn't be that slow

0x72D0:
yeah, but i'm pretty sure we execute the avx of the code. I try to print a word when AVX is executed and it works only for the AVX version of the code. So maybe is an instruction like the proble with addw in neon or it's I implement a bad algorithm.

ihhub:
@0x72D0 could you please measure timing for float based calculations? I have a feeling that there is some performance drawback on AVX functions for double. I'll modify performance test code little later for this.

ihhub:
It could be not a bad implementation of your code but slow realisation of instructions on your CPU. If it's true then we could try to run performance trers on our servers to see results.

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,258 @@
...
+            const float * PointEndSimd = point + totalSimdWidth;
+
+            float * distanceVal = distance.data();
+
+            const __m256 coeff = _mm256_set_ps(cosVal, sinVal, cosVal, sinVal, cosVal, sinVal, cosVal, sinVal);
+            const __m256i ctrl = _mm256_set_epi32(7, 6, 3, 2, 5, 4, 1, 0);
+
+            for( ;point != PointEndSimd; point += avx_float, distanceVal += avx_float )
+            {
+                __m256 src1 = _mm256_loadu_ps(point);
```
ihhub: I think you try to cast it to __m256 * and just use post increment (a++) here and for next load function. What do you think? No performance difference but little nicer code.

ihhub:
The problem might lies in this warning (which is treated as error on Windows/AppVeyor build):
```
c:\projects\penguinv\src\math\hough_transform_simd.cpp(67): warning C4752: found Intel(R) Advanced Vector Extensions; consider using /arch:AVX [C:\projects\penguinv\build\test\unit_tests\unit_tests.vcxproj]
  c:\projects\penguinv\src\math\hough_transform_simd.cpp(114): warning C4752: found Intel(R) Advanced Vector Extensions; consider using /arch:AVX [C:\projects\penguinv\build\test\unit_tests\unit_tests.vcxproj]
```
Searching the internet would give an answer about switching SSE and AVX but I don't

ihhub:
Going to [Intel Website](https://software.intel.com/en-us/node/524134) and it says `No corresponding Intel® AVX instruction.`

ihhub:
@0x72D0 could you please try to use `_mm256_load_ps ` and `_mm256_load_pd ` instead? Just create a fixed array as `const float[8] = {your data}` or `const double[4] = {your data}` and use above instructions.

ihhub:
Okay, I fixed that warning. Now we have failed unit tests on Windows. Waiting for your reply regarding performance tests ;)

0x72D0:
this is the performance I have now:
screenshot_20190102_083536

ihhub:
@0x72D0 Okay, I fixed unit and performance test code. There was a problem that you set simd during test adding into framework while during an execution there was no guarantee that you data would be set correctly. You have to do this within a test itself. Please try again :)

0x72D0:
no, still bad:
screenshot_20190102_084425

ihhub:
@0x72D0 then you have to debug it what's wrong with the code as I am almost out of ideas :) I hope that we would have passed unit tests with my latest changes.

ihhub:
How about `float`? Have you tried to measure timings?

ihhub:
@0x72D0 , you should have a look into latency of instructions. Some of instructions for double calculation have latency as 3-5 which is pretty slow while latency for `float` operations are slower meaning it's faster. That's why I asked to compare `float` first.

ihhub:
Unit tests still fail sometimes :( I can take a look only tomorrow (my tomorrow).

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    template <>
+    void FindDistanceSimd< float >( const std::vector< PointBase2D< float > > & input, std::vector < float > & distance, float cosVal, float sinVal, const size_t inputPointCount )
+    {
+        simd::SIMDType simdType = simd::actualSimdType();
+
+#ifdef PENGUINV_AVX_SET
+        if ( simdType == simd::avx_function && float_layout_check ) {
+            const size_t simdWidth = (inputPointCount*2) / (avx_float*2);
```
ihhub: Here we could remove 2 from both sides of equation.

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,277 @@
...
+    const size_t neon_float = 4u;
+
+    const float minimumAngleStep = 0.001f * static_cast<float>( pvmath::pi ) / 180.0f;
+    const float minimumLineTolerance = 1e-5f;
+    
+    template <typename _Type>
+    void FindDistance( const std::vector< PointBase2D< _Type > > & input, std::vector < _Type > & distance, _Type cosVal, _Type sinVal, size_t offset = 0u )
+    {
+        if ( offset >= input.size() )
+            return;
```
ihhub: Here we better throw `imageException` exception instead so we could catch some logic problem during unit tests run.

src/math/hough_transform_simd.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+#ifdef PENGUINV_SSE_SET
+#include <emmintrin.h>
+#endif
+
+#ifdef PENGUINV_NEON_SET
+#include <arm_neon.h>
+#endif
+
+#include <algorithm>
```
ihhub: Put please these inclusions just after `#include "hough_transform_simd.h"` line as per practice it is recommended to put std library inclusions first.

0x72D0:
the float version of the code is not better:
screenshot_20190102_131716

0x72D0:
ok I think the instruction with the most latency is the multiplication. If I refers to the instruction table of agner fog: https://www.agner.org/optimize/instruction_tables.pdf. the two multiplication is around 5 lantency each, the hadd is around 5 latency and the permutation is 3 lantency.

ihhub:
@0x72D0 thanks for the efforts! I'll take a look little later. We need to analyze the code to first fix unit tests, and secondly find a better way for SIMD implementation.

ihhub:
How the algorithm works:
1. We have these values `[x1, y1, x2, y2], [x3, y3, x4, y4]` loaded into 2 simd variables
2. multiple them by coefficients, result is `[a1, b1, a2, b2], [a3, b3, a4, b4]`
3. do horizontal addition `[a1 + b1, a3 + b3, a2 + b2, a4 + b4]`
4. swap values to be `[a1 + b1, a2 + b2, a3 + b3, a4 + b4]`

What we can do currently is to modify step 4 which has latency of 3. My suggestion is to use 2 function calls instead of 1 and these functions are `_mm256_bslli_epi128 ` and `_mm256_blend_pd`. According to manuals these functions have latency 1 which in result gives us total latency 2. How it works:
```cpp
src1 = _mm256_bslli_epi128( result, 8 ); // put a2 + b2 on second place
result = _mm256_blend_pd( result, src1, 4 + 2);
```

0x72D0:
the performance issue might also be a problem with my processor? Maybe the latency of certain instruction is higher on my processor model?

ihhub:
Yes, it could be. Let's try to optimise the code. I'm thinking how to even remove horizontal add by using shift operations for example. We have to do 2 multiplications for sure as we have 2 variables.

0x72D0:
I tried to find the cause to the random fail, so I ran valgrind memcheck to see if it was a memory leaks and it's not...
screenshot_20190103_164150

0x72D0:
what I find weird about the random fail it's that the double and the float version always fail at the same time...

ihhub:
Try to run 10000 times just these 2 unit tests (comment the rest to be executed). It seems like a random generator related problem.

0x72D0:
ok, I put the testRunCount to 100001 and I got some weird result:
screenshot_20190103_190058

ihhub:
> ok, I put the testRunCount to 100001 and I got some weird result:
> screenshot_20190103_190058

That's very interesting... I'll take a look into line intersection later. Hough transform seems to fail too. Could you please try to change `const _Type noiseValue = lineTolerance / 2;` into `const _Type noiseValue = lineTolerance / 3;` and see if it still fails?

0x72D0:
here's my result after changing the noise value:
screenshot_20190103_214409

and here's all my cpu information:
screenshot_20190103_213851

0x72D0:
when I change the noiseValue, only the houghTransform without simd was slower

ihhub:
> when I changed the noiseValue, only the houghTransform without simd was slower
`noiseValue` is for unit tests, not for performance tests.

Your CPU fully supports AVX. I think you have to debug (set break point on `return false` in unit test) to catch the conditions of failure or just wait for me to do it :)

ihhub:
Line intersection test should be fixed from now.

ihhub:
I ran unit tests on my Windows machine with all changes related to Hough transform for original code with 1,000,000 test loop and no errors appear.

0x72D0:
ok I tried the new line intersection and now it's fixed. But when I try on the original version on linux, hough transform still fail:
screenshot_20190104_110259

0x72D0:
if I put 1,000,001 instead of 100,000 I get different failure condition:
angle : 1.23918
angleTolerance : 0.00349066
angleStep : 0.000120368
lineTolerance : 2.92

ihhub:
There are not only these values of randomisation: an array size and points are randomised too. Could you confirm that during failure you have `pointOffLine` size as 1? If yes you could set break point in `hough_transform.cpp`, line `outOffLine.push_back( (*point) );` and the distance with min and max values as well.
I tried to set given by you values. None of them fails.

ihhub:
@0x72D0 and to shrink the problem: could you please  debug it on latest trunk version without your changes?

0x72D0:
I test on the latest trunk and I found a pointOffLine size of 2

ihhub:
@0x72D0 is it possible to post here all data for hough transform input parameters including all points and angle parameters with line parameters? If I have all data I can easily find the root of problem.

0x72D0:
points:
[292.15, 431.578];[196.751, 292.316];[-305.786, -454.587];[459.913, 683.091];[-28.8128, -43.0271];[102.559, 152.981];[-213.387, -314.188];[-39.2398, -55.6933];[488.085, 723.306];[-492.639, -729.747];[-455.387, -677];[359.105, 535.498];[-219.199, -325.596];[180.376, 269.591];[-262.677, -387.573];[-492.8, -730.296];[-544.735, -806.982];[553.972, 819.125];[102.371, 151.772];[-366.022, -542.03];[-380.799, -562.696];[20.724, 33.517];[143.987, 210.677];[51.8574, 75.951];[-154.868, -230.221];[-304.445, -452.29];[148.492, 218.597];[-510.217, -755.498];[-90.6084, -134.022];[-516.982, -766.767];[147.192, 216.049];[-424.519, -630.616];[-70.7024, -105.441];[345.518, 513.183];[-391.82, -582.448];[481.663, 711.302];[558.341, 824.983];[368.874, 544.706];[-216.844, -321.174];[317.271, 471.614];[-381.561, -568.48];[207.488, 310.096];[120.583, 178.771];[160.307, 239.837];[-47.0737, -69.7896];[-388.488, -575.648];[-148.855, -219.756];[-428.733, -636.244];[347.948, 518.958];[-505.192, -749.909];

angle:
-2.16421

angleTolerance:
0.0191986

angleStep:
0.000548532

lineTolerance:
3.47

pointOnline: 
[292.15, 431.578];[196.751, 292.316];[-305.786, -454.587];[459.913, 683.091];[-28.8128, -43.0271];[102.559, 152.981];[-213.387, -314.188];[-39.2398, -55.6933];[488.085, 723.306];[-492.639, -729.747];[-455.387, -677];[-219.199, -325.596];[180.376, 269.591];[-262.677, -387.573];[-492.8, -730.296];[-544.735, -806.982];[553.972, 819.125];[102.371, 151.772];[-366.022, -542.03];[-380.799, -562.696];[20.724, 33.517];[143.987, 210.677];[51.8574, 75.951];[-154.868, -230.221];[-304.445, -452.29];[148.492, 218.597];[-510.217, -755.498];[-90.6084, -134.022];[-516.982, -766.767];[147.192, 216.049];[-424.519, -630.616];[-70.7024, -105.441];[345.518, 513.183];[-391.82, -582.448];[481.663, 711.302];[558.341, 824.983];[368.874, 544.706];[-216.844, -321.174];[317.271, 471.614];[207.488, 310.096];[120.583, 178.771];[160.307, 239.837];[-47.0737, -69.7896];[-388.488, -575.648];[-148.855, -219.756];[-428.733, -636.244];[347.948, 518.958];[-505.192, -749.909];

pointOffline: 
[359.105, 535.498];[-381.561, -568.48];

ihhub:
It's very strange: my `angleVal` at the end of hough function is `2.16640425`, `bestAngleId` is `-4`, `minDistance` is `-3.51301455`, `maxDistance` is `3.42698550`. Distance value (from `distanceToLine` array) for `[359.105, 535.498]` point (position 11) is `-3.15097046`, `[-381.561, -568.48];` (position 40) `3.06503296`, which are in range.

0x72D0:
it's only happen in the float function, so maybe it's only a question of precision

ihhub:
> it's only happen in the float function, so maybe it's only a question of precision

I'm thinking that something to do with `gcc` compiler, especially with `-ffast-math` flag (or it's equivalent). Could you check whether you set it (by default) or set some precise math flag in this case?

0x72D0:
these seem to be my default flag on gcc: '-mtune=generic' '-march=x86-64'

ihhub:
@0x72D0 I think I might found a bug in the algorithm. `const _Type currentDistance = (distanceToLine[initialPointId + onLinePointCount - 1u] + distanceToLine[initialPointId]) / 2;` seems to be incorrect.

ihhub:
Correction for above statement: it's correct :)

ihhub:
I'm proposing a plan what to do next:
1. Please add needed files in `test/performance_tests/performance_tests.vcxproj` file.
2. Remove all code related to **NEON** and **SSE** as of now. I'd like to concentrate on **AVX** as from my thinking **NEON** and **SSE** are not faster than **AVX**
3. Remove `result = _mm256_permutevar8x32_ps(result, ctrl);` and `result = _mm256_permute4x64_pd(result, 0b11011000);` lines. We know that we sort values after but replace the call `FindDistanceSimd<_Type>(input, distanceToLine, cosVal, sinVal);` (on line 248) with `FindDistance<_Type>(input, distanceToLine, cosVal, sinVal);`. So we would use SIMD only for searching angle inside the loop. At least with removed **permute** function the code would be faster. Please measure results after it.
4. Next step would be to reduce duplication code. If we made above changes only the code inside loop to find distances is different. In this situation I'm proposing to add function pointer definitions inside `hough_transform.h` file (yes, original). These pointers should be to `FindDistanceSimd` functions (only 2 in our case) and make this pointer as an input parameter in original hough transform function with default value as `mullptr`. Then inside `hough_transform.cpp` file inside the loop before finding the distance check if this pointer is not nullptr then call it but if it is nullptr then execute the original code. So in `hough_transform_simd.cpp` you would have only 2 functions. You call original hough transform functions with given function pointers within this cpp file.

ihhub:
@0x72D0 if even after removal of `permute` function AVX code would be slower we might consider to do not go... Could you please check with NEON code latencies? Assume that we use similar functions without `permute`

ihhub:
```
[1/4] HoughTransformFloat... 12.96+/-0.6228 ms
[2/4] HoughTransformDouble... 12.54+/-0.403 ms
[3/4] HoughTransformAvxFloat... 13.54+/-0.5548 ms
[4/4] HoughTransformAvxDouble... 12.74+/-0.3208 ms
```
These are times without `permute` function. As you can see even here AVX is slower than CPU code meaning that we shouldn't waste our time on this anymore as it's not possible to speed up after. Do you agree with this?

0x72D0:
yes I agree, let's drop this issue for now. Maybe we can retry with avx512 in the future

Issue #306, opened 2018-12-25
Histogram calculation with mask


ihhub:
We have an implementation of **Histogram** function in `src/image_function.h` and `src/image_function.cpp` files. This function is pretty simple. 

We want to add the same function but with slightly different logic: function accepts an extra parameter - mask which is actually another binary image. Mask could be as a part of another image as well. We know that Histogram function works with image ROI (region of interest). We read first top-left pixel from mask ROI (which should be the same by size as original image ROI). If pixel intensity if greater than 0 then we read a top-left pixel from original image and put result into histogram. If pixel intensity on mask is equal to 0 then we don't read pixel intensity from original image and we don't put the value into histogram.

There are function definitions for new functions:
```cpp
    std::vector < uint32_t > Histogram( const Image & image, const Image & mask );
    void                     Histogram( const Image & image, const Image & mask, std::vector < uint32_t > & histogram );
    std::vector < uint32_t > Histogram( const Image & image, uint32_t x, uint32_t y, const Image & mask, uint32_t maskX, uint32_t maskY, uint32_t width, uint32_t height );
    void                     Histogram( const Image & image, uint32_t x, uint32_t y, const Image & mask, uint32_t maskX, uint32_t maskY, uint32_t width, uint32_t height, std::vector < uint32_t > & histogram );
```

ihhub:
@yannlabou would you like to take this? This shouldn't take so much time.

PR #325, opened 2019-01-06
Histogram calculation with mask

yannlabou:
Related to #306

### DIFF
--- a/src/image_function.cpp
+++ b/src/image_function.cpp
@@ -776,7 +776,7 @@ namespace Image_Function
         const uint32_t rowSizeMask = mask.rowSize();
 
         const uint8_t * imageY     = image.data() + y * rowSize + x;
-        const uint8_t * imageYMask = mask.data() + yMask * rowSizeMask + xMask;
+        const uint8_t * imageYMask = mask.data() + maskY * rowSizeMask + maskX;
         const uint8_t * imageYEnd  = imageY + height * rowSize;
 
         for( ; imageY != imageYEnd; imageY += rowSize, imageYMask += rowSizeMask ) {
--- a/src/image_function.cpp
+++ b/src/image_function.cpp
@@ -747,6 +747,30 @@ namespace Image_Function
         }
     }
 
+    std::vector < uint32_t > Histogram( const Image & image, const Image & mask )
+    {
+        const Image maskedImage = BitwiseAnd( image, mask );
+        return Image_Function_Helper::Histogram( Histogram, maskedImage );
+    }
+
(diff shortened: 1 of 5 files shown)


src/image_function.cpp
```diff
@@ -747,6 +747,30 @@ namespace Image_Function
         }
     }
 
+    std::vector < uint32_t > Histogram( const Image & image, const Image & mask )
+    {
+        const Image maskedImage = BitwiseAnd( image, mask );
```
ihhub: The idea behind mask is to do not calculate some areas of image, meaning to do not include results from these areas of image. If we use bitwise logical AND the number of results in histogram remains the same. What we need is to add pixel values from original image only on areas where mask has non 0 values. For example, we found a circle on an image and we want a histogram only this circle so we use mask for this.

ihhub: The code in new function should be something like this:
```cpp
        const uint32_t rowSize = image.rowSize();
        const uint32_t rowSizeMask = mask.rowSize();

        const uint8_t * imageY     = image.data() + y * rowSize + x;
        const uint8_t * imageYMask = mask.data() + yMask * rowSizeMask + xMask;
        const uint8_t * imageYEnd  = imageY + height * rowSize;

        for( ; imageY != imageYEnd; imageY += rowSize, imageYMask += rowSizeMask ) {
            const uint8_t * imageX     = imageY;
            const uint8_t * imageXMask = imageYMask;
            const uint8_t * imageXEnd  = imageX + width;

            for( ; imageX != imageXEnd; ++imageX, ++imageXMask ) {
                if( (*imageXMask ) > 0 )
                    ++histogram[*imageX];
            }
        }
```

ihhub:
@yannlabou shall I change the code based on my suggestions or wait for you? Additionally, we need unit tests for this function.

yannlabou:
Feel free if you need it. I'm unfortunately on business trip and short on time this week.

ihhub:
> Feel free if you need it. I'm unfortunately on business trip and short on time this week.

Got it. Have a fruitful trip! ;)

ihhub:
Base changes are done. We need to add unit tests for this. To simplify work I'm proposing to add unit tests in such way: generate uniform image. Generate mask image. In mask image make a half of the image white (255) and another half black (0). Then the total result would be the same as in original Histogram but the histogram would have in 2 times less points.

yannlabou:
@ihhub Thank you for finishing the PR. I wish I had more time these days.

ihhub:
That's fine. It's still better than nothing ;)

Issue #117, opened 2018-10-08
SIMD implementation of RgbToBgr() function


ihhub:
We do not have SIMD (SSE, AVX or NEON) implementation of **RgbToBgr()** function. For the person who is willing to add such code please take a look into an implementation of the function inside `image_function.cpp` and the structure which is used for SIMD code inside `image_function_simd.cpp`. Please do not forget to add unit test for newly created function(s) inside `unit_test_image_function.cpp` (it's just one line code via macros call).

ihhub:
Cool!

ihhub:
We don't have any unit tests for **RgbToBgr** function. I'll think how to make a proper test for it.

0x72D0:
ok no problem

ihhub:
@0x72D0 now you have privileges as a collaborator to edit issues, assign them, change labels and review pull requests :)

0x72D0:
ok the SSE function is ready for testing :)

ihhub:
Great! Looking forward for a pull request.

ihhub:
@0x72D0 for AVX I'm proposing a way to use shift operator by 16 bits (**_mm256_sll_epi32**) and then use blend function **_mm256_blendv_epi8**. What do you think?

ihhub:
Ah, it won't work that way :( We need to figure out based on shifts...

0x72D0:
why `_mm256_blendv_epi8` won't work, if we do _mm256_blendv_epi8(result,result,ctrl)?

ihhub:
First of all you need to have all of your B values on first place. Currently they are on third. You have have values in simd: RGBRGBRGB... (I didn't write all). Now to put all B values on R position we could use 16 bit shift but variables are 32 bit so we have: RGBR GBRG. As you can see first 32 bit integer should be shifted by 16 bits, while second cannot be shifted by such amount (out of range).

0x72D0:
I still didn't found an algorithm for AVX... I think I will go back to this issue later, maybe I'm gonna find more idea

PR #334, opened 2019-01-11
SSE implementation of RgbToBgr() function

0x72D0:
the SSE version is ready to test, only the unit_test for RgbToBgr is missing
relates to #117

### DIFF
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -944,7 +944,7 @@ namespace sse
     }
 
     void RgbToBgr( uint8_t * outY, const uint8_t * inY, const uint8_t * outYEnd, uint32_t rowSizeOut, uint32_t rowSizeIn, 
-                   const uint8_t colorCount, uint32_t simdSize, uint32_t totalSimdWidth, uint32_t nonSimdWidth )
+                   const uint8_t colorCount, uint32_t simdWidth, uint32_t totalSimdWidth, uint32_t nonSimdWidth )
     {
         const simd ctrl = _mm_set_epi8(15, 12, 13, 14, 9, 10, 11, 6, 7, 8, 3, 4, 5, 0, 1, 2);
         for( ; outY != outYEnd; outY += rowSizeOut, inY += rowSizeIn ) {
@@ -953,7 +953,7 @@ namespace sse
 
             const uint8_t * outXEnd = outX + totalSimdWidth;
 
-            for( ; outX != outXEnd; outX += simdSize, inX += simdSize ) {
+            for( ; outX != outXEnd; outX += simdWidth, inX += simdWidth ) {
                 const simd * src = reinterpret_cast<const simd*> (inX);
                 simd * dst = reinterpret_cast<simd*> (outX);
                 simd result = _mm_loadu_si128( src );
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -34,6 +34,7 @@ namespace
             table.Maximum            = &Image_Function_Simd::Maximum;
             table.Minimum            = &Image_Function_Simd::Minimum;
             table.ProjectionProfile  = &Image_Function_Simd::ProjectionProfile;
+            table.RgbToBgr           = &Image_Function_Simd::RgbToBgr;
             table.Subtract           = &Image_Function_Simd::Subtract;
             table.Sum                = &Image_Function_Simd::Sum;
             table.Threshold          = &Image_Function_Simd::Threshold;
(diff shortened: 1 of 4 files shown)


src/image_function_simd.cpp
```diff
@@ -941,6 +942,35 @@ namespace sse
         }
     }
 
+    void RgbToBgr( uint8_t * outY, const uint8_t * inY, const uint8_t * outYEnd, uint32_t rowSizeOut, uint32_t rowSizeIn, uint32_t RGB_simdSize, 
+                   const uint8_t colorCount, uint32_t simdWidth, uint32_t totalSimdWidth, uint32_t nonSimdWidth )
+    {
+        simd ctrl = _mm_set_epi8(15, 12, 13, 14, 9, 10, 11, 6, 7, 8, 3, 4, 5, 0, 1, 2);
```
ihhub: `const simd`

src/image_function_simd.cpp
```diff
@@ -941,6 +942,35 @@ namespace sse
...
+            const uint8_t * inX  = inY;
+            uint8_t       * outX = outY;
+
+            const uint8_t * outXEnd = outX + totalSimdWidth;
+
+            for( ; outX != outXEnd; outX += RGB_simdSize, inX += RGB_simdSize ) {
+                const simd * src = reinterpret_cast<const simd*> (inX);
+                simd * dst = reinterpret_cast<simd*> (outX);
+                simd result = _mm_loadu_si128( src );
+                result = _mm_shuffle_epi8(result, ctrl);
```
ihhub: [MacOS compilation](https://travis-ci.org/ihhub/penguinV/jobs/478124648) fails on this function. Check the bottom of the log.

ihhub: It's probably because this function is SSSE3.

src/image_function_simd.cpp
```diff
@@ -1935,6 +1965,49 @@ if ( simdType == neon_function ) { \
...
+        const uint32_t rowSizeOut = out.rowSize();
+
+        width = width * colorCount;
+
+        const uint8_t * inY  = in.data()  + startYIn  * rowSizeIn  + startXIn  * colorCount;
+        uint8_t       * outY = out.data() + startYOut * rowSizeOut + startXOut * colorCount;
+
+        const uint8_t * outYEnd = outY + height * rowSizeOut;
+
+        const uint32_t RGB_simdSize = (simdSize/3)*3;
```
ihhub: `const uint32_t RGBSimdSize = (simdSize / 3) * 3;`

src/image_function_simd.cpp
```diff
@@ -1935,6 +1965,49 @@ if ( simdType == neon_function ) { \
...
+
+        uint32_t simdWidth = width / RGB_simdSize;
+        uint32_t totalSimdWidth = simdWidth * RGB_simdSize;
+        uint32_t nonSimdWidth = width - totalSimdWidth;
+
+        // to prevent unallowed access to memory
+        if(nonSimdWidth < (simdSize%3))
```
ihhub: if( nonSimdWidth < (simdSize % 3) )

src/image_function_simd.cpp
ihhub: This function is declared in `tmmintrin.h` file which is not included here. Check [this](https://software.intel.com/en-us/node/524215)

src/image_function_simd.cpp
```diff
@@ -941,6 +942,35 @@ namespace sse
         }
     }
 
+    void RgbToBgr( uint8_t * outY, const uint8_t * inY, const uint8_t * outYEnd, uint32_t rowSizeOut, uint32_t rowSizeIn, uint32_t RGB_simdSize,
```
ihhub: Use `simdWidth` variable instead of `RGB_simdSize` to keep the same style.

0x72D0:
ok now we just need on a correct unit_test for this function

ihhub:
I'm working on it right now.

ihhub:
@0x72D0 could you please add performance tests for this pull request while unit test pull request is compiled?

ihhub:
Unit tests are committed. Please merge and add a line to include unit tests for SSE case.

0x72D0:
the performance is a little bit better

screenshot_20190112_120947

ihhub:
@0x72D0 a small recommendation: please during test run to do execute any other applications giving full CPU usage to tests. Sigma value which comes after average value should be pretty low, <5% than average value. Other applications affect on test results if they eat a lot of CPU resources. If you do such you would have more accurate results so you could check performance enhancement more accurate ;)

0x72D0:
oh okay I understand, I did it again with only the performance test running
screenshot_20190112_134440

ihhub:
@0x72D0, thank you very much for implementing this!

PR #341, opened 2019-01-16
Remove duplicate code in unit and performance tests

MeKot:
This is still a WIP and need a few touch ups and some cleaning


ihhub:
@MeKot could you please your branch with main branch of the library. I see that your pulll request includes changes which you made for previous pull request. That's one of the source of file conflicts.

ihhub:
If you need a help in merging please don't resitate to ask. I would like to help in any occured problem to faster solve the issue :)

test/performance_tests/CMakeLists.txt
```diff
@@ -10,6 +11,8 @@ set(SRCS
     ${LIB_DIR}/image_function_helper.cpp
     ${LIB_DIR}/image_function_simd.cpp
     ${LIB_DIR}/penguinv/penguinv.cpp
+    ${TEST_LIB}/test_helper.h
```
ihhub: No need to add header files into CMake file.

test/performance_tests/performance_test_helper.cpp
```diff
@@ -133,6 +133,7 @@ namespace Performance_Test
         return timer.mean();
     }
 
+    /*
```
ihhub: We don't support an usage of multiline comments as it's hard to see what's really commented. Please remove the code instead.

test/test_helper.cpp
```diff
@@ -1,18 +1,17 @@
-#include "unit_test_helper.h"
```
ihhub: The initial idea was to just move some functions into a specific file. Now this file has an extra set of functions which are not needed for performance tests. Could you please put back functions which are not common in `unit_test_helper.cpp` back? I'm sorry for such extra move

test/performance_tests/CMakeLists.txt
```diff
@@ -10,6 +11,8 @@ set(SRCS
...
+    ${TEST_LIB}/test_helper.cpp
```
ihhub: Let's just use here `../test_helper.cpp` code.

ihhub:
Hi @MeKot , how is it going? If you stuck or need a help just ask me please. Me could discuss before making more changes. Alternatively, you could pick up any other task which is simpler and then come back to this issue later :)

ihhub:
I got an idea: why don't we move (as it's intended originally) only similar functions under Test_Unit namespace and then inside existed namespaces just put `using namespace Test_Unit` leading to no changes in most of the files?

MeKot:
I'll restart with that immediately and it should be done in a few hours.

ihhub:
Sure. Thank you very much!

Issue #115, opened 2018-10-08
SIMD implementation of ConvertToRgb() function


ihhub:
We do not have SIMD (SSE, AVX or NEON) implementation of **ConvertToRgb()** function. For the person who is willing to add such code please take a look into an implementation of the function inside `image_function.cpp` and the structure which is used for SIMD code inside `image_function_simd.cpp`. Please do not forget to add unit test for newly created function(s) inside `unit_test_image_function.cpp` (it's just one line code via macros call).

0x72D0:
I'm gonna try this one next

ihhub:
Sure!

ihhub:
@0x72D0 how is it going? If you need any help, I'd like to discuss :)

0x72D0:
oh I didn't got the time to start it yet, I should start it tonight

ihhub:
It would be great to see the implementation soon :)

ihhub:
Hi @ihhub are you going to implement NEON code as well? :)

ihhub:
Hi @0x72D0  are you going to implement NEON code as well? :)

0x72D0:
yeah sure :). I am not sure how to implement the AVX one since the shuffle is a little bit weird...

ihhub:
Don't worry about AVX as of now. It's better to have something than trying to have everything ;) An idea for AVX would come later.

ihhub:
@0x72D0 did you have a time for NEON implementation? :)

0x72D0:
not yet, I might have some time for this, this weekend.

ihhub:
Closing this issue as AVX implementation is not obvious and we do not have a proper solution.

PR #342, opened 2019-01-20
SSE implementation of ConvertToRgb function

0x72D0:
PR for #115 

Here's the performance:
screenshot_20190120_163854


ihhub:
@0x72D0 thank you very much for your great work!

PR #343, opened 2019-01-21
Remove duplicate code in tests  

MeKot:
relates to #341


test/unit_tests/unit_test_helper.h
```diff
@@ -10,17 +10,11 @@ namespace Unit_Test
     // Generate images
```
ihhub: In this file I was proposing to add something like `using namespace Test_Helper` so all changes for unit_test_image_function.cpp for example won't be needed ;)

MeKot: I see.. Would you like me to revert those and place a `using namespace` instead? :)

ihhub: Would be nice of course. My apology that you spent so much time but I see that my idea would cost less changes in code changes after all... What is your opinion?

MeKot: Most of the time I prefer to specify the namespace of the function to avoid confusing the compiler/completion engine and whoever reads the code, but, since we're extracting all of the functions into a single namespace, it makes perfect sense to have `using`. No worries of the time, in all honesty, I should have finished it ages ago

MeKot: @ihhub I've refactored out the use of `Test_Helper namespace` references from ``unit_test_helper``. If you'd like me to do any more changes - do tell :)

ihhub: Mmm... what I meant is this:
1. In file `unit_test_helper.h` add `using namespace Test_Helper;` at the beginning of `namespace Unit_Test`. We don't need to put it inside source file.
2. The same should be done for `performance_test_helper.h`: put `using namespace Test_Helper;` at the beginning of `namespace Performance_Test`
3. Revert all changes for:
```cpp
- performance_test_blob_detection.cpp
- performance_test_filtering.cpp
- performance_test_image_function.cpp
- unit_test_image_function_cuda.cpp
- unit_test_image_function_opencl.cpp
- unit_test_blob_detection.cpp
- unit_test_edge_detection.cpp
- unit_test_fft.cpp
- unit_test_image_buffer.cpp
- unit_test_image_function.cpp
- unit_test_math.cpp
```

In result you would use Unit_Test and Performance_Test as some kind of extension of Test_Helper namespace. Now it's a combination of 2 different namespaces which is confusing for other developers.

Does it make more sense to do such?

ihhub:
By the way @MeKot would you like to work on other issues after the completion of this one? We have a couple of issues similar to this for code duplication reduction or would you like to pick up something else if you have more interest in another areas?

test/unit_tests/unit_test_helper.h
MeKot: Ooooooh, yes, that makes so much more sense to me now! I'll take care of this right away

MeKot:
I'd love to! It would be nice to complete an issue on duplication somewhere in `src` (to get familiar with the code there) and then move on onto other types of issues.

test/unit_tests/unit_test_helper.h
ihhub: I'll have a look little later but from first view everything looks good.

ihhub:
@MeKot the only difference in uniformImage functions between unit tests and performance tests is that for performance tests we specify width and height while for unit tests we don't. Why don't we try to merge these functions making width and height as 2 prlast parameters before reference? Default values for them would be 0.  In function itself we check if any parameter is 0 we use random generated value otherwise intial value.

ihhub:
@MeKot thanks a lot for fixing this issue!

PR #351, opened 2019-01-26
neon implementation of ConvertToRgb function

0x72D0:
PR for #350 and #115.

all unit test pass:

screenshot_20190126_110056

Here's the performance test:

screenshot_20190126_110443

### DIFF
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -1325,6 +1325,46 @@ namespace neon
         }
     }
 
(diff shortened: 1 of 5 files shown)


ihhub:
Thank you! I'll have a look on it at my morning. Just a quick note: multithreading version is in 5 times slower than neon version while neon has 20x speed up from CPU code. Something might be wrong.

0x72D0:
maybe is a problem with my particular CPU? Maybe you can try on an other CPU?

ihhub:
@0x72D0 thanks a lot for NEON implementation of this function. I don't have ARM based hardware right now but I'll keep in mind this issue.

Issue #358, opened 2019-01-28
Add missing function registration for ConvertToRgb SIMD function


ihhub:
In`src\image_function_simd.cpp` we have a registration structure **FunctionRegistrator** where we register functions to be used everywhere across the library. We forgot to add `ConvertToRgb` function.

ihhub:
@0x72D0 could you please add this function and run unit and performance tests for it? Now function pool shouldn't have 5x slower performance than SIMD :)

0x72D0:
sorry... I completly forgot about it...

PR #361, opened 2019-01-28
Add missing function registration for ConvertToRgb SIMD function

0x72D0:
PR for #358. here's the performance on SSE:

screenshot_20190128_091109

I'm gonna upload the result for NEON later when I'm gonna have access to my pi.


ihhub:
That's interesting. Does function pool work faster now?

ihhub:
@0x72D0 thank you very much for fixing this issue.
Could you please post your results from Raspberry Pi here? It's not urgent, just want to verify that we don't have 5x gap between multithreaded and simd versions.

PR #365, opened 2019-02-02
Travis builds Python Module with Python3

yannlabou:
Python 3.7 is not available on Ubuntu 14.04 (Trusty). Using Python 3.6.


ihhub:
@yannlabou thanks a lot of fixing this issue.

Issue #348, opened 2019-01-26
Python module compilation for Python 2 (for Python 3 is done) on AppVeyor


ihhub:
We have auto compilation of Python module for Python 3.7 on AppVeyor. We want to add auto compilation for Python 2.7 as well as many people still use this version of Python. What changes we need:
1. Rename **PythonModule** jobs into **Python3.7**
2. Add new jobs named **Python2.7**
3. Add the same logic for Python2.7 as it's done for Python3.7

yannlabou:
It was difficult but I submitted /pull/374.

ihhub:
@yannlabou thank you so much for your efforts! I put few comments there :)

PR #374, opened 2019-02-07
Build Python Module on Appveyor with Python 2.7

yannlabou:
relates to #348

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,6 @@ configuration:
 environment:
   matrix:
     - PYTHON: "C:\\Python27-x64"
-      WINDOWS_SDK_VERSION: "v7.1"
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
     - PYTHON: "C:\\Python37-x64"
 
@@ -35,6 +34,7 @@ for:
   matrix:
     only:
       - configuration: PythonModule
+  platform: Any CPU
   init:
     - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
     - python --version
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,7 +34,9 @@ for:
   matrix:
     only:
       - configuration: PythonModule
-        platform: x64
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -46,7 +53,7 @@ before_build:
   - cmd: if "%configuration%"=="Release" call script\appveyor\cuda_install.cmd
 
 build_script:
-  - set PENGUINV_UNIT_TEST_RUN_COUNT=500
+  - set PENGUINV_UNIT_TEST_RUN_COUNT=5
```
ihhub: Do we need thia change? I think we needed this for testings only :)

appveyor.yml
```diff
@@ -11,33 +11,40 @@ version: 1.0.{build}
...
+  matrix:
+    - JOB: Release_x86
+      configuration: Release
+      platform: x86
+    - JOB: Release_x64
+      configuration: Release
+      platform: x64
+    - JOB: Python2.7
+      PYTHON: "C:\\Python27-x64"
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
```
ihhub: Do we need this change?

ihhub: Am I correct if I say that this is needed for `call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86_amd64` line?

appveyor.yml
yannlabou: This was for testing obviously.

appveyor.yml
yannlabou: Yes. On Appveyor, Python2 otherwise always invoke MSVC++ 9 (Visual Studio 2008 version 9.0). `SET VS90COMNTOOLS=%VS140COMNTOOLS%` just doesn't help here. APPVEYOR_BUILD_WORKER_IMAGE is unfortunately undocumented.

ihhub:
@yannlabou thanks a lot for addinf this feature!

Issue #376, opened 2019-02-11
Add support for Mingw compiler


BascoDLowell:
I'm trying to compile the sample project in examples/qt_framework/qimage_display, and I got 2 errors stating '_cpuindex' was not declared in scope. I'm compiling in Qt.

I am also interested in contributing code (the reason I'm trying the example), can someone help me started?

ihhub:
Hello @BascoDLowell could you please paste here exact compilation errors? It would greatly help to identify the root of problem.

ihhub:
Which are would you like to contribute in? We have UI, GPU, SIMD, code optimisation, code cleanup or even documentation? :)

ihhub:
@BascoDLowell plus could you please state your OS as well and development environment?

BascoDLowell:
![penguinerrorimage](https://user-images.githubusercontent.com/14083887/52552368-e6570200-2e1a-11e9-9e2f-ea4d041a8547.JPG)

ihhub:
I assume you use Windows. Give me please 2-3 hours until I get an access to Windows machine to check the issue. Thank you for sharing it!

BascoDLowell:
My OS is Windows 7. I use Qt 5.6 with mingw compiler.

ihhub:
I think the problem lies in the compiler. It might not support such functions...

ihhub:
@BascoDLowell could you please try to use another compiler to verify such problem?

BascoDLowell:
I have an installer for Qt 5.7 with built-in msvc 2015 compiler. I'll try that one.

ihhub:
Please try this trick: in `src/penguinv/cpu_identification.h` file change line 6 `#ifdef _WIN32 // Windows` into `#if defined(_WIN32) && !defined (__MINGW32__)`.
Currently workflow was tested only for Visual Studio compiler on Windows.

BascoDLowell:
Sorry, I lost my previous qt installation. I also cannot run the example with Qt 5.7.1 with MSVC2015 compiler.I got this message:
Error while building/deploying project example_qimage_display (kit: Desktop Qt 5.7.1 MSVC2015_64bit)
When executing step "qmake"

BascoDLowell:
I'll get back to this tomorrow.

ihhub:
Okay. Let's wait for your tomorrow's test :)

ihhub:
[...]

PR #377, opened 2019-02-12
Add MinGW compiler support

BascoDLowell:
Fix Issue #376 .


ihhub:
We need to wait for our compilation server completion to make sure that these changes won't break any compilation. After that I'll merge the code.

ihhub:
@BascoDLowell thank you very much for fixing this issue!

Issue #282, opened 2018-12-02
Add circle (oval) and rectangle drawing support for UI Window classes


ihhub:
We have 3 classes for UI Window:
`src/ui/ui.h` - base class
`src/ui/qt/qt_ui.h` - QT API based class
`src/ui/win/win_ui.h` - Win32 API based class

Currently classes support drawings of points and lines, we need to add functions to support drawings of **ovals** and **rectangles** as well.

This issue relates to #104

ihhub:
Hi @BascoDLowell, would you like to have a look into this issue since you work on QT (and on Windows)? You could make changes for QT as of now (and then later fix for Windows). If you need more information regarding this issue or something is no clear I would gladly help you :)

BascoDLowell:
I'll try this out @ihhub . I like to ask:
Do we have an example for the drawPoint() and drawLine() function in Qt?
This functions should draw to a bitmap image, right? (changing colors in the bitmap stream?)

ihhub:
We already have an implementation of point and line functions for Qt. Please take a look into **src/ui/qt/qt_ui.h** and **src/ui/qt/qt_ui.cpp** files. We use paint tool to draw on canvas and then output it as a bitmap image.

BascoDLowell:
I'm interested on digging on this! I don't have much knowledge on drawing in QT but I have time to read documentations.

ihhub:
I would recommend to have a look into [QPainter documentation](https://doc.qt.io/qt-5/qpainter.html). For ellipse you could use **drawEllipse** while for rectangle you could use a set of **drawLine** calls. I hope it would simplify your work :)

BascoDLowell:
Hey @ihhub ! I finished extending the qt_ui.h classes with ellipse and rectangle. Should I open a Pull Request? or work on Win32 API additional classes?
Can I still use Qt IDE with Win32  API? I'm not sure I know how to add Win32 API to a project. Can you help me?

ihhub:
You could create a pull request for just Qt only if you think it's easy for you. No harm in this.

You can use Win32 API within Qt as this API is OS based API. If I'm wrong you could have a look at Win32 UI example in `examples\windows\edge_detection`.

BascoDLowell:
I noticed when I tried creating a pull request that my fork didn't have your commit when you changed the comments a while ago, can you help me regarding that? How can I add that commit in my fork?

ihhub:
[...]

PR #378, opened 2019-02-12
Add oval and rectangle drawing support for UI Window Qt class

BascoDLowell:
**`drawEllipse` function** - implemented from `QPainter::drawEllipse( int x,int y,int w, int h)` which refers to a rectangle with starting point (x,y), width w, and height h)
**Parameters:**
`Point2D center` - starting point referring to center of ellipse.
`int xDiameter` - Diameter in x axis, 
`int yDiameter `- Diameter in y axis
`PaintColor color` - color of drawn ellipse

**`drawRectangle` function**
**Parameters:**
`Point2D start` - starting point referring to upper left point of rectangle.
`int width` - width from starting point to its right direction.
`int height` - height from starting point to its bottom direction.
`PaintColor color` - color of drawn rectangle

Related to Issue #282

### DIFF
--- a/src/ui/qt/qt_ui.cpp
+++ b/src/ui/qt/qt_ui.cpp
@@ -45,13 +45,11 @@ void UiWindowQt::drawLine( const Point2d & start, const Point2d & end, const Pai
     _display();
 }
 
-void UiWindowQt::drawEllipse( const Point2d & center, double xDiameter, double yDiameter, const PaintColor & color )
+void UiWindowQt::drawEllipse( const Point2d & center, double xRadius, double yRadius, const PaintColor & color )
 {
     QPainter paint( &_pixmap );
     paint.setPen( QColor( color.red, color.green, color.blue, color.alpha ) );
-    paint.drawEllipse( static_cast<int>(center.x - (xDiameter / 2)), static_cast<int>(center.y - (yDiameter / 2)),
-                       static_cast<int>(xDiameter), static_cast<int>(yDiameter) );
-
+    paint.drawEllipse( QPointF(center.x, center.y), xRadius, yRadius );
     _window.setPixmap( _pixmap );
     _display();
 }
--- a/src/ui/qt/qt_ui.cpp
+++ b/src/ui/qt/qt_ui.cpp
@@ -45,21 +45,21 @@ void UiWindowQt::drawLine( const Point2d & start, const Point2d & end, const Pai
     _display();
 }
 
-void UiWindowQt::drawEllipse(const Point2d & center, const int xDiameter, const int yDiameter, const PaintColor &color)
+void UiWindowQt::drawEllipse(const Point2d & center, double xDiameter, double yDiameter, const PaintColor &color)
 {
     QPainter paint( &_pixmap );
     paint.setPen( QColor( color.red, color.green, color.blue, color.alpha ) );
-    paint.drawEllipse(static_cast<int>(center.x)-(xDiameter/2),static_cast<int>(center.y)-(yDiameter/2),xDiameter,yDiameter);
+    paint.drawEllipse(static_cast<int>((center.x)-(xDiameter/2)),static_cast<int>((center.y)-(yDiameter/2)),static_cast<int>(xDiameter),static_cast<int>(yDiameter));
 
     _window.setPixmap( _pixmap );
     _display();
 }
 
-void UiWindowQt::drawRectangle(const Point2d & start, const int width, const int height, const PaintColor &color)
+void UiWindowQt::drawRectangle(const Point2d & topLeftCorner, double width, double height, const PaintColor &color)
 {
     QPainter paint( &_pixmap );
     paint.setPen( QColor( color.red, color.green, color.blue, color.alpha ) );
-    paint.drawRect(static_cast<int>(start.x),static_cast<int>(start.y),width,height);
+    paint.drawRect(static_cast<int>(topLeftCorner.x),static_cast<int>(topLeftCorner.y),static_cast<int>(width),static_cast<int>(height));
 
     _window.setPixmap( _pixmap );
     _display();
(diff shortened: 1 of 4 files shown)


src/ui/qt/qt_ui.h
```diff
@@ -12,6 +12,8 @@ class UiWindowQt : public UiWindow
 
     virtual void drawPoint( const Point2d & point, const PaintColor & color );
     virtual void drawLine( const Point2d & start, const Point2d & end, const PaintColor & color );
+    virtual void drawEllipse(const Point2d & center, const int xDiameter, const int yDiameter, const PaintColor & color);
```
ihhub: You make these functions as *virtual* but base class doesn't have them. We must add such functions in base class as well.

src/ui/qt/qt_ui.cpp
```diff
@@ -44,3 +44,23 @@ void UiWindowQt::drawLine( const Point2d & start, const Point2d & end, const Pai
     _window.setPixmap( _pixmap );
     _display();
 }
+
+void UiWindowQt::drawEllipse(const Point2d & center, const int xDiameter, const int yDiameter, const PaintColor &color)
```
ihhub: Diameter should be in `double`.

src/ui/qt/qt_ui.cpp
```diff
@@ -44,3 +44,23 @@ void UiWindowQt::drawLine( const Point2d & start, const Point2d & end, const Pai
...
+{
+    QPainter paint( &_pixmap );
+    paint.setPen( QColor( color.red, color.green, color.blue, color.alpha ) );
+    paint.drawEllipse(static_cast<int>(center.x)-(xDiameter/2),static_cast<int>(center.y)-(yDiameter/2),xDiameter,yDiameter);
+
+    _window.setPixmap( _pixmap );
+    _display();
+}
+
+void UiWindowQt::drawRectangle(const Point2d & start, const int width, const int height, const PaintColor &color)
```
ihhub: Remove `const` modifier from base variable types such as `int`.

src/ui/qt/qt_ui.h
```diff
@@ -12,6 +12,8 @@ class UiWindowQt : public UiWindow
...
+    virtual void drawRectangle(const Point2d & start, const int width, const int height, const PaintColor & color);
```
ihhub: I recommend to use 4 `Point2d` points instead.

BascoDLowell: Is this for the rectangle only, or for ellipse as well? I think it is easier to understand drawEllipse() if parameters follow geometric convention. What do you think?

ihhub: It's only for Rectangle.

BascoDLowell:
Thanks for prompt suggestions. I'll continue this tomorrow.

src/ui/qt/qt_ui.h
BascoDLowell: I think we could just add a separate function that accepts 4 `Point2d`, which could draw other quadrilaterals aside from rectangle. We could call it `drawQuadrangle()` or `drawQuadrilateral` or `drawTetragon()`. `drawRectangle()` would be inaccurate name if we accept 4 `Point2d`. What do you think?

ihhub: Having only one point is not enough as user doesn't know which corner of rectangle it represents. Additionally, there is no restriction top (or bottom) line of rectangle must be parallel to X axis (in other words rectangle must no be rotated).
Let's make like this to avoid misunderstanding:
```cpp
virtual void drawRectangle(const Point2d & topLeftCorner, double width, double height, const PaintColor & color);
```
`Point2d` is **double** based structure.

BascoDLowell: Alright! :+1: I will implement the changes after I finished reformatting my computer.

src/ui/qt/qt_ui.cpp
```diff
@@ -44,3 +44,25 @@ void UiWindowQt::drawLine( const Point2d & start, const Point2d & end, const Pai
     _window.setPixmap( _pixmap );
     _display();
 }
+
+void UiWindowQt::drawEllipse( const Point2d & center, double xDiameter, double yDiameter, const PaintColor & color )
+{
+    QPainter paint( &_pixmap );
+    paint.setPen( QColor( color.red, color.green, color.blue, color.alpha ) );
+    paint.drawEllipse( static_cast<int>(center.x - (xDiameter / 2)), static_cast<int>(center.y - (yDiameter / 2)),
```
ihhub: Why don't we use `drawEllipse(const QPointF & center, qreal rx, qreal ry)` instead? It seems more reasonable. Plus we can use **radius** instead of **diameter** which brings more sense.

BascoDLowell: I'm ok with radius..
But regarding the datatype QPointF and qreal, I'm thinking:
This is Qt specific data types which is fine for qt_ui but how will the base class be affected if we use these datatypes? Will it be fine to use different datatypes in virtual functions in our base class and the derived class?

ihhub: My apology for late reply. What I meant is to change only Qt drawEllipse function without changing our function interface. We can do casting from double to float if QPointF is float based class/structure. Leaving accuracy in such case is not a big problem as drawings are for human eye and person cannot recognise very small subpixel shift.

ihhub: `QPointF` accepts two `qreal` variables which are `double` so our code would look like:
```cpp
paint.drawEllipse( QPointF(center.x, center.y), xRadius, yRadius );
```
Without any changes in our function's input parameter list :)

BascoDLowell: Thanks!
I'll implement this in my next commit 👍

ihhub:
@BascoDLowell thank you very much for your contribution to the project!

We're looking forward for your Win32 API implementation ;)

PR #386, opened 2019-02-18
Add oval and rectangle drawing support for Win32 API based UI Window Class

BascoDLowell:
Rectangle was implemented using 4 lines.
Ellipse was implemented using Arc() function. 

Related to Issue #282

### DIFF
--- a/src/ui/win/win_ui.cpp
+++ b/src/ui/win/win_ui.cpp
@@ -283,9 +283,9 @@ void UiWindowWin::drawEllipse( const Point2d & center, double xRadius, double yR
 
 void UiWindowWin::drawRectangle( const Point2d & topLeftCorner, double width, double height, const PaintColor & color )
 {
-    const Point2d topRightCorner = Point2d( topLeftCorner.x + width, topLeftCorner.y );
-    const Point2d bottomLeftCorner = Point2d( topLeftCorner.x, topLeftCorner.y + height );
-    const Point2d bottomRightCorner = Point2d( topLeftCorner.x + width, topLeftCorner.y + height );
+    const Point2d topRightCorner( topLeftCorner.x + width, topLeftCorner.y );
+    const Point2d bottomLeftCorner( topLeftCorner.x, topLeftCorner.y + height );
+    const Point2d bottomRightCorner( topLeftCorner.x + width, topLeftCorner.y + height );
 
     _line.emplace_back( topLeftCorner, topRightCorner, color );
     _line.emplace_back( topLeftCorner, bottomLeftCorner, color );
--- a/src/ui/win/win_ui.cpp
+++ b/src/ui/win/win_ui.cpp
@@ -68,14 +68,14 @@ namespace WindowsUi
             }
 
             for ( std::vector < UiWindowWin::EllipseToDraw >::const_iterator ellipse = _window->_ellipse.cbegin(); ellipse != _window->_ellipse.cend(); ++ellipse ) {
-                const int _xTopLeft = static_cast<int>( ellipse->xTopLeft * xFactor);
-                const int _yTopLeft = static_cast<int>( ellipse->yTopLeft * yFactor);
-                const int _xBottomRight = static_cast<int>( ellipse->xBottomRight * xFactor);
-                const int _yBottomRight = static_cast<int>( ellipse->yBottomRight * yFactor);
+                const int xTopLeft = static_cast<int>( ellipse->topLeft.x * xFactor);
+                const int yTopLeft = static_cast<int>( ellipse->topLeft.y * yFactor);
+                const int xBottomRight = static_cast<int>( ellipse->bottomRight.x * xFactor);
+                const int yBottomRight = static_cast<int>( ellipse->bottomRight.y * yFactor);
 
                 HPEN hPen = CreatePen( PS_SOLID, 1, RGB( ellipse->color.red, ellipse->color.green, ellipse->color.blue ) );
                 HGDIOBJ hOldPen = SelectObject( hdc, hPen );
-                Arc( hdc, _xTopLeft, _yTopLeft, _xBottomRight, _yBottomRight, 0, 0, 0, 0 );
+                Arc( hdc, xTopLeft, yTopLeft, xBottomRight, yBottomRight, 0, 0, 0, 0 );
                 SelectObject( hdc, hOldPen );
                 DeleteObject( hPen );
             }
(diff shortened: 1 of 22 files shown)


ihhub:
Could you please remove line 14 in the example (`main.cpp` file) as well which I mentioned?

BascoDLowell:
Done removing the line 👍

src/ui/win/win_ui.cpp
```diff
@@ -66,6 +66,19 @@ namespace WindowsUi
                 SelectObject( hdc, hOldPen );
                 DeleteObject( hPen );
             }
+
```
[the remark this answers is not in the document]
BascoDLowell: Oh, that's why the indentation doesn't look as good here as in Visual Studio. I'll take note of that!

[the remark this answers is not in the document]
ihhub: You could change settings in Visual Studio to use 4 spaces instead of tab for indentation.

BascoDLowell:
@ihhub, the code revisions were ready for your review. Please tell me if I missed something. 👍

src/ui/win/win_ui.h
```diff
@@ -52,8 +54,27 @@ class UiWindowWin : public UiWindow
...
+        EllipseToDraw( double xTopLeft_, double yTopLeft_, double xBottomRight_, double yBottomRight_, const PaintColor & color_ = PaintColor() )
+            : xTopLeft(xTopLeft_)
+            , yTopLeft(yTopLeft_)
+            , xBottomRight(xBottomRight_)
+            , yBottomRight(yBottomRight_)
+            , color(color_)
+        {
+        }
+
+        double xTopLeft;
```
ihhub: What if we name them as:
```cpp
double left;
double top;
double right;
double bottom;
```
? :)

src/ui/win/win_ui.cpp
```diff
@@ -66,6 +66,19 @@ namespace WindowsUi
...
+            for ( std::vector < UiWindowWin::EllipseToDraw >::const_iterator ellipse = _window->_ellipse.cbegin(); ellipse != _window->_ellipse.cend(); ++ellipse ) {
+                const int _xTopLeft = static_cast<int>( ellipse->xTopLeft * xFactor);
```
ihhub: No need to add **_** for local variables. We do this only for members of a class.

src/ui/win/win_ui.h
BascoDLowell: xTopLeft and yTopLeft refers to one point while xBottomRight and yBottomRight is another point.

ihhub: Then use `Point2d` structure for it.

src/ui/win/win_ui.h
```diff
@@ -52,8 +54,27 @@ class UiWindowWin : public UiWindow
         PaintColor color;
     };
 
+    struct EllipseToDraw
+    {
+        EllipseToDraw( double xTopLeft_, double yTopLeft_, double xBottomRight_, double yBottomRight_, const PaintColor & color_ = PaintColor() )
```
ihhub: Add please default values for variables.

BascoDLowell:
@ihhub, please review the changes. Thanks!

src/ui/win/win_ui.cpp
```diff
@@ -66,6 +66,19 @@ namespace WindowsUi
...
+
+            for ( std::vector < UiWindowWin::EllipseToDraw >::const_iterator ellipse = _window->_ellipse.cbegin(); ellipse != _window->_ellipse.cend(); ++ellipse ) {
+                const int xTopLeft = static_cast<int>( ellipse->topLeft.x * xFactor);
+                const int yTopLeft = static_cast<int>( ellipse->topLeft.y * yFactor);
+                const int xBottomRight = static_cast<int>( ellipse->bottomRight.x * xFactor);
+                const int yBottomRight = static_cast<int>( ellipse->bottomRight.y * yFactor);
+
+                HPEN hPen = CreatePen( PS_SOLID, 1, RGB( ellipse->color.red, ellipse->color.green, ellipse->color.blue ) );
+                HGDIOBJ hOldPen = SelectObject( hdc, hPen );
+                Arc( hdc, xTopLeft, yTopLeft, xBottomRight, yBottomRight, 0, 0, 0, 0 );
```
ihhub: What if we use [Ellipse](https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/nf-wingdi-ellipse) function instead? Then we just need to name variables as `left, right, top, bottom`.

BascoDLowell: I've read the documentation. The naming of the function is misleading. But essentially, they are the same.
image

ihhub: In my opinion we should use `Ellipse` function as it has less arguments and doesn't rise questions why we need to put so many zeros in function parameters :)

Regarding the naming: x coordinate of top left corner is left position, y coordinate of top left corner if top position. Naming a variable as `xTopLeft` is equal to `left` but the second case doesn't have excessive data in names. That's what I mean :) In most of image processing applications people use 4 sides instead of corners.

BascoDLowell: So this refers to the sides. I see now.. Ok I'll be changing this in my next commit. 👍

BascoDLowell: image
I got this from using Ellipse.

BascoDLowell: image
This is Arc.

ihhub: Then in this case you're totally right :)

src/ui/win/win_ui.cpp
```diff
@@ -261,6 +274,27 @@ void UiWindowWin::drawLine( const Point2d & start, const Point2d & end, const Pa
...
+void UiWindowWin::drawEllipse( const Point2d & center, double xRadius, double yRadius, const PaintColor & color )
+{
+    _ellipse.emplace_back( center.x - xRadius, center.y - yRadius, center.x + xRadius, center.y + yRadius, color );
+
+    _display();
+}
+
+void UiWindowWin::drawRectangle( const Point2d & topLeftCorner, double width, double height, const PaintColor & color )
+{
+    const Point2d topRightCorner = Point2d( topLeftCorner.x + width, topLeftCorner.y );
```
ihhub: You can use a constructor here instead of assign operator. **const Point2d name(x, y);**

BascoDLowell:
I made the change in taking the four sides in EllipseToDraw. 👍 I'm willing to work on other issues if this gets merged, please advise me :)

ihhub:
Sure. I'll do merging soon. I'll advice you what you couls try to do next ;)

src/ui/win/win_ui.cpp
BascoDLowell: I'm not sure I get what you mean.

ihhub: I'll make changes. You'll see :)

ihhub:
@BascoDLowell thank you very much for implementing this feature!

Issue #357, opened 2019-01-28
SIMD implementation of Flip() function 


ihhub:
We do not have SIMD (SSE, AVX or NEON) implementation of **Flip()** function. For the person who is willing to add such code please take a look into an implementation of the function inside `image_function.cpp` and the structure which is used for SIMD code inside `image_function_simd.cpp`. Please do not forget to add unit test for newly created function(s) inside `unit_test_image_function.cpp` (it's just one line code via macros call).

ihhub:
Hi @0x72D0 , would you like to have a look into this issue? I think we could implement it for horizontal flip at least which is still better than nothing. What do you think?

ihhub:
Hello @0x72D0 do you have a spare time for have a look into this? :)

0x72D0:
I'm gonna check this this weekend :)

ihhub:
Sure. Looking forward for the implementation!

0x72D0:
with my test, I get the following error:
![screenshot_2019-02-17_11-45-57](https://user-images.githubusercontent.com/16087328/52916126-cd7e9e80-32a9-11e9-8e71-52f438785b80.png)
any idea why this error trigger?

ihhub:
Original Flip function works only with gray-scale images so the problem could lie in VerifyColorImage instead of correct VerifyGrayScaleImage usage.

ihhub:
You just post a code for `Flip` function from `namespace simd` which you made and I take a look :)

0x72D0:
from the simd namespace:

![screenshot_2019-02-17_22-49-55](https://user-images.githubusercontent.com/16087328/52927030-81157c00-3306-11e9-84b0-613c35666fd1.png)

from the sse namespace (for now I try to make the horizontal only working):

![screenshot_2019-02-17_22-50-29](https://user-images.githubusercontent.com/16087328/52927059-a0aca480-3306-11e9-9714-3f232d6a5f2a.png)

ihhub:
You call `ConvertToRgb` inside code which is wrong.

0x72D0:
oh yeah you're right... didn't see it...

0x72D0:
but still it shouldn't be the problem since when I test sse, these lines are not executed

ihhub:
Unit tests generate image which could have width less than 16 bytes leading to execution of CPU path code.

0x72D0:
oh ok, I will ran the test and see if the output change

[...]

PR #387, opened 2019-02-18
SSE implementation of Flip function

0x72D0:
PR for #357 

performance when horizontal = true, vertical = false:

screenshot_2019-02-17_23-26-57

performance when horizontal = true, vertical = true:

screenshot_2019-02-17_23-28-48


ihhub:
@0x72D0 thank you very much for this implementation! I hope you would have a time to do the same on NEON :)

PR #388, opened 2019-02-19
Testing on Code-Style Branch

yannlabou:


ihhub:
Hi @yannlabou is it correct to say that not all obvious formatting errors are caught?

yannlabou:
> Hi @yannlabou is it correct to say that not all obvious formatting errors are caught?

What do you mean? At first I wanted to see that the script is running for pull requests. This works. Now we still need to create a format file. This the toughest part IMHO.

ihhub:
Sorry, my bad. Could you please point to an example of format file. I'll review and write here what we want to do. Then you will review too and then we add it :)

yannlabou:
> Sorry, my bad. Could you please point to an example of format file. I'll review and write here what we want to do. Then you will review too and then we add it :)

No need to be sorry. I'm actually learning much more from you than you from me :).
I pushed a `.clang-format` file I have been using for personal projects. We can use it at start and iterate. You had probably want things like no space between ìf`and parenthesis etc.

yannlabou:
[clang-format-3.9 doc](http://releases.llvm.org/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html)

ihhub:
Thanks fo the example! I'll comment if I have some thoughts...

.clang-format
```diff
@@ -0,0 +1,72 @@
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: true
+AlignEscapedNewlinesLeft: Right
```
ihhub: I don't get that does this option do :(

.clang-format
```diff
@@ -0,0 +1,72 @@
...
+    BeforeCatch: true
+    BeforeElse: true
+    SplitEmptyFunction: false
+    SplitEmptyRecord: false
+    SplitEmptyNamespace: false
+BreakBeforeBinaryOperators: All
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: BeforeColon
+BreakStringLiterals: false
+ColumnLimit: 100
```
ihhub: 150-170 for now. Let's set 170.

.clang-format
```diff
@@ -0,0 +1,72 @@
...
+CompactNamespaces: true
```
ihhub: I cannot find this option in the manual.

.clang-format
```diff
@@ -0,0 +1,72 @@
...
+IndentPPDirectives: AfterHash
+IndentWidth: 4
+IndentWrappedFunctionNames: true
+KeepEmptyLinesAtTheStartOfBlocks: false
+Language: Cpp
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: Inner
+PenaltyBreakBeforeFirstCallParameter: 19937
+PenaltyReturnTypeOnItsOwnLine: 19937
+PointerAlignment: Left
```
ihhub: Middle.

.clang-format
```diff
@@ -0,0 +1,72 @@
...
+BreakStringLiterals: false
+ColumnLimit: 100
+CompactNamespaces: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 0
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+FixNamespaceComments: true
+IncludeBlocks: Preserve
```
ihhub: I cannot find this option :(

.clang-format
```diff
@@ -0,0 +1,72 @@
...
+IndentCaseLabels: false
+IndentPPDirectives: AfterHash
+IndentWidth: 4
+IndentWrappedFunctionNames: true
+KeepEmptyLinesAtTheStartOfBlocks: false
+Language: Cpp
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: Inner
+PenaltyBreakBeforeFirstCallParameter: 19937
```
ihhub: What does this number mean? :)

.clang-format
```diff
@@ -0,0 +1,72 @@
...
+CompactNamespaces: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 0
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+FixNamespaceComments: true
+IncludeBlocks: Preserve
+IndentCaseLabels: false
+IndentPPDirectives: AfterHash
```
ihhub: I cannot find this option :(

.clang-format
```diff
@@ -0,0 +1,72 @@
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
```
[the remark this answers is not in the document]
yannlabou: You're the boss :)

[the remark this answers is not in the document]
ihhub: Hey, I wrote my explanation only :) If you have another opinion I would be glad to hear ;)

.clang-format
yannlabou: Something like this.
```
#define A   \
  int aaaa; \
  int b;    \
  int dddddddddd;
```

.clang-format
yannlabou: You're right. Not available for clang-format-3.9.

.clang-format
yannlabou: Not available for clang-format-3.9.

.clang-format
yannlabou: Not available for clang-format-3.9.

.clang-format
ihhub: What if we have something like:
```cpp
bool a;
const uint8_t * b;
```
For this case alignment would be weird :)

yannlabou: It's only about the escape `\` symbol like in macros or comments.

ihhub: Then it's totally fine. We use alignment for macros like this.

.clang-format
yannlabou: A random huge number :). It's to be able to break the line length rule.
```
Namespaces::Are::Pervasive::SomeReallyVerySuperDuperLongFunctionName(args);
// and the column limit is here:                                        ^
```
clang-format will probably do something like this:
```
Namespaces::Are::Pervasive::SomeReallyVerySuperDuperLongFunctionName(
    args);
```

.clang-format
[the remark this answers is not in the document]
yannlabou: I'm ok. Those are really not that important. We won't never agree about `PointerAlignment`. ;)

.clang-format
yannlabou: Done.

ihhub:
@yannlabou I got a good idea!

What if we create for example `.yamllint` for CodeFactor? CodeFactor is static analysis tool which we use. It has a support of `Yamllint, ShellCheck, CppLint, Pylint`?

yannlabou:
> @yannlabou I got a good idea!
> 
> What if we create for example `.yamllint` file for CodeFactor? CodeFactor is static analysis tool which we use. It has a support of `Yamllint, ShellCheck, CppLint, Pylint`?

I can't really follow you. What yaml files do you want to lint? `cpplint` is also an option but it has a fairly good support only for the `Google` style.

ihhub:
Sorry for mis clarification. What I mean is that `CodeFactor` is a web site which can include multiple static analyzers. We can add `.yamllint` file for **Yamllint** checks within *CodeFactor*. Of course we cannot cover overall but we could visually see problems in the code if any.

Could you please revert changes for `examples/bitmap_operation/example_bitmap_operation.cpp` file so that I can merge this pull request> :)

yannlabou:
OK I see. I have no experience with codefactor. Could you please create the initial `.yamllint` file to demonstrate what you mean?

I will revert later today. I’m on my mobile.

ihhub:
Sure. I will do it my tomorrow :)

ihhub:
@yannlabou thank a lot for your work!

By the way you could create a separate branch within this repository so you could easily use compilation remote servers as well.

yannlabou:
> By the way you could create a separate branch within this repository so you could easily use compilation remote servers as well.

This is a great idea :). Thank you!

PR #390, opened 2019-02-21
NEON implementation of Flip function

0x72D0:
PR for #357 

performance of horizontal = true and vertical = true:

screenshot_2019-02-21_12-01-20

performance of horizontal = true vertical = false:

screenshot_2019-02-21_12-04-21


ihhub:
@0x72D0 thank you very much for your continuous contribution to the project!

Issue #280, opened 2018-11-28
Add Format Step into Build Pipeline


yannlabou:
To help maintainers and contributors, a style guideline should be provided and directly ran on Appveyor and Travis CI. We can ensure that all changes are checked for formatting before they are merged into master. [clang-format](http://clang.llvm.org/docs/ClangFormat.html) has been established in the industry as standard tool for code formatting.

ihhub:
I totally agree with this. I wanted to to this long time ago. First of all we need to write a coding style so that people could refer to it. Next step is to implement this coding into script verification tools.

yannlabou:
> I totally agree with this. I wanted to to this long time ago. First of all we need to write a coding style so that people could refer to it. Next step is to implement this coding into script verification tools.

Please create a branch for this issue. We could then draft our ideas in Contributing.md.

ihhub:
Shouldn't it be a separate document? Do you have an access to wiki pages of the project?

yannlabou:
> Shouldn't it be a separate document? Do you have an access to wiki pages of the project?

Sure. A separate document linked in the Contributing page. The wiki might also be a great place for that. Anyway let’s start. :)

ihhub:
[Here](https://github.com/ihhub/penguinV/wiki/Code-style). Can you edit it?

yannlabou:
> [Here](https://github.com/ihhub/penguinV/wiki/Code-style). Can you edit it?

Yes.

ihhub:
Hi @yannlabou , you seem to be very busy these days, don't you? :)

Would you mind to share some ideas of implementation for this issue? Such feature greatly improves pull request reviews and overall code development.

yannlabou:
> Hi @yannlabou , you seem to be very busy these days, don't you? :)
> 
Oy yeah. I'm in a new project at work. Lot of stuff to organize. Little time to code :(
> Would you mind to share some ideas of implementation for this issue? Such feature greatly improves pull request reviews and overall code development.
Sure :).

ihhub:
I think we need to close this issue. We need only one time to check format style which is done in Travis-CI.

PR #395, opened 2019-02-27
Add Format Step into Build Pipeline

ihhub:
relates to #280


ihhub:
@yannlabou I was wrong regarding `Yamllint`. It's only for **YAML** files. I'm sorry for delaying this issue.

Issue #100, opened 2018-10-02
Add OpenCL types as it's done for CUDA


ihhub:
For CUDA we have our own implementation of base structures for data handling which locate in `src/cuda/cuda_types.cuh` file. These template classes are `Type` (for a single variable) and `Array` for an array holder. We would like to add similar classes for OpenCL memory as well.

PR #396, opened 2019-02-27
Add base OpenCL types support

ihhub:
relates to #100

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -60,7 +60,7 @@ SpaceBeforeAssignmentOperators: true
 SpaceBeforeParens: ControlStatements
 SpaceInEmptyParentheses: false
 SpacesBeforeTrailingComments: 1
-SpacesInAngles: false
+SpacesInAngles: true
 SpacesInCStyleCastParentheses: false
 SpacesInParentheses: true
 SpacesInSquareBrackets: false
--- a/.clang-format
+++ b/.clang-format
@@ -47,7 +47,7 @@ IndentWrappedFunctionNames: true
 KeepEmptyLinesAtTheStartOfBlocks: false
 Language: Cpp
 MaxEmptyLinesToKeep: 1
-NamespaceIndentation: Inner
+NamespaceIndentation: All
 PenaltyBreakBeforeFirstCallParameter: 19937
 PenaltyReturnTypeOnItsOwnLine: 19937
 PointerAlignment: Middle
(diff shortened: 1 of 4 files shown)


ihhub:
@yannlabou here is a good example of static analysis to be corrected. First thing what I see is that there is no indentation in namespace (proposed solution has no 4 whitespaces shift).

yannlabou:
> @yannlabou here is a good example of static analysis to be corrected. First thing what I see is that there is no indentation in namespace (proposed solution has no 4 whitespaces shift).

I think `NamespaceIndentation: All` is what you need.

ihhub:
Yeah, done this and some other changes as well :) Thanks!

yannlabou:
> Yeah, done this and some other changes as well :) Thanks!

I think we should mention clang-format in [CONTRIBUTING.md](https://github.com/ihhub/penguinV/blob/master/CONTRIBUTING.md).

ihhub:
Moreover, I'm thinking that we could remove [code style](https://github.com/ihhub/penguinV/wiki/Code-Style) page and just mention to refer to this file. From another side not many people would read and understand `clang-format` file.

yannlabou:
> Moreover, I'm thinking that we could remove [code style](https://github.com/ihhub/penguinV/wiki/Code-Style) page and just mention to refer to this file. From another side not many people would read and understand `clang-format` file.

Yes. The page is not needed anymore. Contributors don’t need to read or understand .clang-format (although it’s really self explanatory). They need to run the tool before submitting a pull request. Many editors have support for it. The hardcore vim users probably know clang ;).

ihhub:
> > Moreover, I'm thinking that we could remove [code style](https://github.com/ihhub/penguinV/wiki/Code-Style) page and just mention to refer to this file. From another side not many people would read and understand `clang-format` file.
> 
> Yes. The page is not needed anymore. Contributors don’t need to read or understand .clang-format (although it’s really self explanatory). They need to run the tool before submitting a pull request. Many editors have support for it. The hardcore vim users probably know clang ;).
My worry is not for hardcore users but for beginners. Let's just mention about it as we always can change the code within pull request (or ask a person to do such).

ihhub:
After all coding style is just a recommendation, not a set of strict rules which we must force to follow :) Even if there is some differences in style we shouldn't make a big accent on it as such act pushes developers away.

Issue #397, opened 2019-03-01
Merge opencl_types.h and cuda_types.cuh files


ihhub:
We have `src/opencl/opencl_types.h` and 'src/cuda/cuda_types.cuh` files. Both of them contain 2 template classes which are almost identical. We want to reduce code duplication for such case. Please create a new file called `src/gpu/base_types.h` which would include a common code for these two files.

manekenpix:
I'd like to work on this one

ihhub:
Hi @manekenpix , sure!

PR #398, opened 2019-03-01
Create base_types.h

ReshmaPD:
create a new file to merge two files containing common code
implement for issue #397


ihhub:
Hi @ReshmaPD thank you for your efforts but we need more than just add an empty file to fix #397 issue. Could you please implement it?

ReshmaPD:
I was confused about this, I should have clarified it first.
Regarding the issue about comparing the common code and creating a new file, can you please elaborate about how should I approach that? Take the common code from both and implement that in a new file? I am a total beginner and would love to try.

ihhub:
You're correct. We need to create a common classes for both files.

Please put new base classes under `namespace gpu` in your newly created file. These new classes must be templates.
The biggest problem is that OpenCL implementation doesn't allow copy constructor (it's private). Don't worry if you don't have CUDA or OpenCL compilation. Our remote servers (Travis-CI and AppVeyor) do all needed compilation. Plus you can add as many commits as you need to archive the goal. We will still merge all commits into one :)

If you have any questions regarding the implementation please ask. We could discuss even on half made code if you put it for this pull request.

ihhub:
@ReshmaPD if you are not really familiar with templates (which is fine) I could suggest to try to fix another issue like [this](https://github.com/ihhub/penguinV/issues/366) or [this](https://github.com/ihhub/penguinV/issues/373). These issues are pretty straightforward :)

ihhub:
Hi @ReshmaPD how is the progress?

ihhub:
I'm closing this pull request for now as there are no changes in the code.

Issue #394, opened 2019-02-27
Processing acceleration for full width ROI


ihhub:
Recently (#87 ) we introduced processing acceleration for full width ROI in SIMD functions (`src/image_function_simd.cpp`). We could apply the same technique for CPU functions located in `src/image_function.cpp` file. Steps what we need to do:
1. Move **IsFullImageRow** and **OptimiseRoi** functions into `src/parameter_validation.h` file. We need to use template argument for images instead of predefined `Image` class.
2. Move `limits` inclusion to `src/parameter_validation.h` file.
3. In `src/image_function.cpp` file add **OptimiseRoi** function in each function where it's possible.

This change would speed up functions and reduce unit test execution timing as well.

ihhub:
I'll fix this issue :)

ihhub:
@0x72D0 out change for SIMD doesn't work for color images :) I'll fix it too.

PR #400, opened 2019-03-01
Processing acceleration for full width ROI

ihhub:
relates to #394


0x72D0:
are you sure about putting width = width * colorCount; before?  will it cause trouble for checking the rowSize condition?

ihhub:
Yes.
```
            _rowSize = width() * colorCount();
            if( _rowSize % alignment() != 0 )
                _rowSize = (_rowSize / alignment() + 1) * alignment();
```

0x72D0:
oh ok I see

Issue #406, opened 2019-03-07
Add internal memory copy between CPU and GPU when a new option is enabled


ihhub:
We present the library as a project with heterogeneous system support but we haven't implemented an internal memory copy function within the code. How it should work: assume we have a global option which we could locate in `src/image_function_helper.h` and `src/image_function_helper.cpp` called *bool internalMemoryCopy;*.
A developer has a bunch of classes/functions which are registered within `src/penginv/penguinv.cpp` file. The developer creates an object on GPU and executes functions/classes with this object. Right now if any of registered functions/classes don't have an implementation for GPU we throw an exception. What we need to do is to convert image object from GPU type into CPU type for this function call and return back to GPU type when any of functions/classes occur with registration for GPU.

Issue #340, opened 2019-01-15
Examples review


ihhub:
We have examples of library usage. Some of them seem to be not up to date or don't attend an interest of potential users. We would like to modify examples to be more user friendly. This topic is a common topic for discussions and ideas.

PR #409, opened 2019-03-16
Add Heterogeneous computing support

ihhub:
relates to #406 and #340


ihhub:
Hi @yannlabou . Would you agree to move images to **data** directory to store files non related to source code?

PR #432, opened 2019-04-15
Issue 430: Fixed logic for building examples and tests

yannlabou:
The CMake control option `PENGUINV_BUILD_TEST` is not removed but the logic to opt-in (or out) for building examples and tests has been fixed. We now check if `PENGUINV_BUILD_TEST` is `ON` **and** if the currently processed `CMakeLists.txt` is the most top-level one.


ihhub:
@yannlabou thanks a lot for fixing this issue! Great work as usual :)

Issue #310, opened 2018-12-27
Merge CPU and SIMD code for image functions


ihhub:
Our CPU based code for image functions locates in `src/image_function.h` and `src/image_function.cpp` files, while SIMD based - in `src/image_function_simd.h` and `src/image_function_simd.cpp`. At initial stage of project development these parts of code were separate and independent. Now both depend on `src/image_function_helper.h` file. We would like to merge them together in this case as there is no drawbacks in doing such. All code would be located in `src/image_function.h` and `src/image_function.cpp` files.

This is not urgent issue.

ihhub:
@MeKot I think this task would suit your needs and interest ;) Please let me know if you have any questions. Also I have a design in mind.

ihhub:
1. `image_function_helper.h` has an enumeration `SIMDType`. This is an indicator which type of function would be called. In `image_function_simd.cpp` we already have checks for this and none of SIMD is available we call CPU code.
2. Use `image_function_simd.cpp` code as a basement. Rename namespace into `Image_Function`. Rename file into `image_function.cpp` (it would replace original file). The explanation why to do such is below. So you would have renamed `image_function_simd.cpp` and original `image_function.h` files.
3. We need to add a namespace called `namespace cpu` just before `namespace avx`. All functions would receive input parameters which are already calculated in functions from `namespace simd`. So we need to add a macro to call CPU function just before calculating simd variables:
```cpp
        const uint8_t * outYEnd = outY + height * rowSizeOut;

        CPU_CODE( parameters )

        const uint32_t simdWidth = width / simdSize;
        const uint32_t totalSimdWidth = simdWidth * simdSize;
        const uint32_t nonSimdWidth = width - totalSimdWidth;

        AVX_CODE( parameters )
        SSE_CODE( parameters )
        NEON_CODE( parameters )
```

Put into this namespace only functions which **have** simd implementation.
4. Copy functions which **don't** have SIMD code as it is into `namespace Image_Function` like it's done in original `image_function.cpp` file.
5. Remove `image_function_simd.h` file.

After this we would need changes in example and tests. We could discuss this later once preparation work would be done.

ihhub:
Hi @0x72D0 would you like to do this since you are very familiar with the code?

0x72D0:
Hi @ihhub I have some trouble to keep up with my study, so I'm gonna restart contributing in 1-2 months. When I'm gonna restart contributing, it would be a pleasure for me to do this issue :)

PR #434, opened 2019-04-16
Merge Image_Function and Image_Function_Simd

0x72D0:
PR for #310 

I have some error with the unit_test, every call to simd function fail...

I still have to refactor the performance_test code


ihhub:
Hi @0x72D0 , thanks for your efforts. I'd definitely have a look this week but it would take some time as there are big changes in the code :)

ihhub:
What we need is to modify existing code for CPU: we don't need to have verification and variable calculation code in each function. We should pass all needed already calculated parameters from common function which now calls SIMD functions.

0x72D0:
ok I understand, but before changing this we should find why the SIMD code is not working in this version. Also how I should rename the simd namespace (since it's now contains all the common function)?

ihhub:
Hi @0x72D0 , thanks a lot for efforts. I'll take a deep look into it on my weekends as currently I'm extremely busy :(

0x72D0:
@ihhub no problem :)

ihhub:
Hi @0x72D0 please check code from line 2270 in `src/image_function.cpp` :)

0x72D0:
@ihhub thanks for spotting this :)

ihhub:
Hi @0x72D0 should we hold this changes for a while? There are not critical of course but they take a lot of time.

0x72D0:
Yeah, right now I'm stuck on why all the SIMD function unit_test don't work... I agree with the fact that these changes are not critical. So if you think we better hold on them, I'm with you on this!

ihhub:
We're closing this pull request for now as it takes a big redesign.

Issue #435, opened 2019-04-18
Rename EdgeDetection into EdgeDetectionBase


ihhub:
We have our edge detection code in `src/edge_detection.h` and `src/edge_detection.cpp` files. `EdgeDetection` class is a template class with default typename **double**. From user perspective a creation of such class should be in this way:
```cpp
EdgeDetection detection;
```
while now user should do like this:
```cpp
EdgeDetection<> detection;
```
which is confusing.
We should rename this class into `EdgeDetectionBase` and use `typedef EdgeDetectionBase<double> EdgeDetection;` to avoid such inconvenience for user.

Issue #414, opened 2019-03-26
Sequencial call of find() function for edge detection doesn't clean previous data


ihhub:
Edge detection code which locates in **src/edge_detection.h** and **src/edge_detection.cpp** files has *find()* function. Sequencial call of this function doesn't clean previous data.

ihhub:
Hi @vrozin , since you have an interest to work on edge detection would you mind to look into this issue? It's just to call `clear()` function of 2 `std::vector` members of EdgeDetection class :)

vrozin:
Absolutely :)
Consider it being done. I'll apply the change while working on #346

PR #440, opened 2019-04-23
Rename EdgeDetection and clear data before finding edge points

ihhub:
relates to #414 and #435


ihhub:
Hi @vrozin could you take a look into these changes and comment if any ideas come into your mind? :)

vrozin:
I like the way you implemented it. `typedef EdgeDetectionBase<double> EdgeDetection` does give a better user experience in the `main` indeed.

Issue #427, opened 2019-04-15
Remove duplication code in BlobInfo class


ihhub:
Our blob detection code locates in `src/blob_detection.h` and `src/blob_detection.cpp` files. There are 2 functions which have the same code: **BlobInfo::_getElongation()** and **BlobInfo::_getLength()**. Steps are needed to be done:
1. In **class BlobInfo** add a new parameter called `Value < double >_guideAngle;`
2. In **BlobInfo::_getLength()** function calculate such angle (see the code in **_getElongation** function)
3. In **BlobInfo::_getElongation()** remove existing code to calculate length and angle and replace it with call of **_getLength()** function.

This issue relates to #69

FrankJWH:
If this is still a viable ticket, I'd like to take a shot at it!

FrankJWH:
The one issue I have refactoring this is that both functions are dependent on [_Point2d startPoint_](https://github.com/ihhub/penguinV/blob/d5bc1a3a795c48fa41e60cdfccc79259bb24bc46/src/blob_detection.cpp#L203) as you can see for [here](url) for **BlobInfo::_getElongation** function and [here]() for **BlobInfo::_getLength**.  Do you suggest creating a member variable for it?

ihhub:
Hi @FrankJWH , yes, this issue is still available and it would be great if you fix it.

I have a better idea: we could create a function unnamed namespace (at the top of `blob_detection.cpp` file which would have such signature:
```cpp
double findLengthFromCountour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, double & length Point2DBase< uint32_t > & start, Point2DBase< uint32_t > & end )
{
    // here we put our logic
}
```

The function returns length value. We need to verify input data and throw an exception if the sizes of contours are not equal. Assign (0, 0) to start and end points if the size of contour is less than 2. Please not that **start** and **end** are *uint32_t* based points. We don't need type conversion during cycle runtime like it's done now.

Is it a clear explanation? :)

FrankJWH:
Yes I believe I understand what you're getting at. I was thinking of creating a function and wanted to follow your coding style and form. I'll get that done as soon as possible!

ihhub:
Sure!

PR #446, opened 2019-04-29
Remove duplication code in BlobInfo class

FrankJWH:
Removed the duplicate code from both member functions and creating a helper function too. Hoping this is exactly what you were looking for!

relates to #427

### DIFF
--- a/src/blob_detection.cpp
+++ b/src/blob_detection.cpp
@@ -232,7 +232,7 @@ namespace Blob_Detection
             if( _contourX.size() > 1 ) {
                 PointBase2D< uint32_t > startPoint;
                 PointBase2D< uint32_t > endPoint;
-                const double length = getLengthFromCountour( _contourX, _contourY, _length.value, startPoint, endPoint );
+                const double length = getLengthFromCountour( _contourX, _contourY, startPoint, endPoint );
                 const double angle = -atan2( static_cast<double>(endPoint.y - startPoint.y),
                                              static_cast<double>(endPoint.x - startPoint.x) );
 
@@ -279,7 +279,7 @@ namespace Blob_Detection
         if( !_contourX.empty() && !_contourY.empty() && !_length.found ) {
             PointBase2D< uint32_t > startPoint;
             PointBase2D< uint32_t > endPoint;
-            _length.value = getLengthFromCountour( _contourX, _contourY, _length.value, startPoint, endPoint );
+            _length.value = getLengthFromCountour( _contourX, _contourY, startPoint, endPoint );
 
             _length.found = true;
         }
--- a/src/blob_detection.cpp
+++ b/src/blob_detection.cpp
@@ -16,8 +16,8 @@ namespace
         CONTOUR    = 4u
     };
     
-    double getLengthFromCountour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, PointBase2D< uint32_t > & start,
-                                 PointBase2D< uint32_t > & end )
+    double getLengthFromCountour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, PointBase2D< uint32_t > & startPoint,
+                                 PointBase2D< uint32_t > & endPoint )
     {
         if ( contourX.size() > 1 ) {
             std::vector < uint32_t >::const_iterator x   = contourX.cbegin();
@@ -37,11 +37,11 @@ namespace
                     if ( maximumDistance < distance ) {
                         maximumDistance = distance;
 
-                        start.x = *x;
-                        start.y = *y;
+                        startPoint.x = *x;
+                        startPoint.y = *y;
 
-                        end.x = *xx;
-                        end.y = *xx;
+                        endPoint.x = *xx;
+                        endPoint.y = *xx;
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


ihhub:
Hi @FrankJWH , thanks for your efforts!

Unfortunately, compilation fails for Unix system. We will try to fix it.

src/blob_detection.cpp
```diff
@@ -142,6 +142,43 @@ namespace Blob_Detection
     {
         return _contourX.size() == _edgeX.size();
     }
+    
+    double BlobInfo::_getLengthFromContour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, double & length, Point2d & start, Point2d & end )
```
ihhub: @FrankJWH ,  it is slightly different from what I proposed in original issue as
```cpp
double getLengthFromCountour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, double & length Point2DBase< uint32_t > & start, Point2DBase< uint32_t > & end )
```
Could you please explain why such change is more suitable? :)

src/blob_detection.cpp
```diff
@@ -192,41 +229,13 @@ namespace Blob_Detection
 
     void BlobInfo::_getElongation()
     {
-        if( !_contourX.empty() && !_contourY.empty() && !_elongation.found ) {
+       if( !_contourX.empty() && !_contourY.empty() && !_elongation.found ) {
```
ihhub: Here we removed 1 whitespace, we need to revert back.

src/blob_detection.cpp
```diff
@@ -142,6 +142,43 @@ namespace Blob_Detection
...
+                        end.x = *xx;
+                        end.y = *xx;
+                    }
+                }
+            }
+
+            length = sqrt( static_cast<double>(maximumDistance) );
+            _guideAngle.value = -atan2( static_cast<double>(endPoint.y - startPoint.y),
+                                        static_cast<double>(endPoint.x - startPoint.x) );
+        }else{
```
ihhub:
```cpp
}
else {
```

src/blob_detection.cpp
```diff
@@ -142,6 +142,43 @@ namespace Blob_Detection
...
+                        start.y = *y;
+
+                        end.x = *xx;
+                        end.y = *xx;
+                    }
+                }
+            }
+
+            length = sqrt( static_cast<double>(maximumDistance) );
+            _guideAngle.value = -atan2( static_cast<double>(endPoint.y - startPoint.y),
```
ihhub: Only one of 2 called functions uses angle. Do we really need to make such value as a member of the class?

src/blob_detection.cpp
ihhub: This function returns `double` by definition but there is no code to do such in function's body.

src/blob_detection.cpp
```diff
@@ -142,6 +142,43 @@ namespace Blob_Detection
...
+                        start.x = *x;
+                        start.y = *y;
+
+                        end.x = *xx;
+                        end.y = *xx;
+                    }
+                }
+            }
+
+            length = sqrt( static_cast<double>(maximumDistance) );
```
ihhub: Here we just need to return value instead of assigning to `length`.

src/blob_detection.cpp
```diff
@@ -142,6 +142,43 @@ namespace Blob_Detection
...
+            _guideAngle.value = -atan2( static_cast<double>(endPoint.y - startPoint.y),
+                                        static_cast<double>(endPoint.x - startPoint.x) );
+        }else{
+            length = 0;
```
ihhub: Here we just need to have `return 0`.

src/blob_detection.cpp
```diff
@@ -192,41 +229,13 @@ namespace Blob_Detection
...
-                    }
-                }
-
-                const double length = sqrt( static_cast<double>(maximumDistance) );
-                const double angle  = -atan2( static_cast<double>(endPoint.y - startPoint.y),
-                                              static_cast<double>(endPoint.x - startPoint.x) );
-
-                const double _cos = cos( angle );
-                const double _sin = sin( angle );
+                const double _cos = cos( _guideAngle.value );
```
ihhub: We could actually calculate angle from startPoint and endPoint. I think we don't need angle as a member of the class.

src/blob_detection.cpp
```diff
@@ -192,41 +229,13 @@ namespace Blob_Detection
...
-                    std::vector < uint32_t >::const_iterator xx = x + 1;
-                    std::vector < uint32_t >::const_iterator yy = y + 1;
-
-                    for( ; xx != end; ++xx, ++yy ) {
-                        uint32_t distance = (*x - *xx) * (*x - *xx) + (*y - *yy) * (*y - *yy);
-
-                        if( maximumDistance < distance ) {
-                            maximumDistance = distance;
+            	Point2d startPoint, endPoint;
+            	_getLengthFromContour(_contourX, _contourY, _length.value, startPoint, endPoint);
```
ihhub: Shouldn't it be like this:
```cpp
const double length = _getLengthFromContour(_contourX, _contourY, startPoint, endPoint);
```
?

ihhub:
@FrankJWH thanks for your commitment to the project!

PR #464, opened 2019-05-17
Issue-458: Set mfpu=neon-vfpv4 as CMake compile option

yannlabou:
With CMake there is no proper way to differentiate between Linux flavors. Nevertheless, I think the flag is only relevant for ARM targets. Therefore the provided fix should help. @0x72D0 Would you mind to test?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,10 @@ if(CUDA_FOUND)
     endif()
 endif()
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)")
+    add_compile_options(-mfpu=neon-vfpv4)
+endif()
+
 option(PENGUINV_BUILD_TEST "Build tests of penguinV" ON)
 if(${PENGUINV_BUILD_TEST} AND (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR))
     enable_testing()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ if(CUDA_FOUND)
     endif()
 endif()
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm*|ARM*)")
     add_compile_options(-mfpu=neon-vfpv4)
 endif()
 


ihhub:
Hi @0x72D0 could you please test this change on your Raspberry Pi?

CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ if(CUDA_FOUND)
     endif()
 endif()
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)")
+    add_compile_options(-mfpu=neon-vfpv4)
```
ihhub: I have a doubt that Visual Studio supports such compiler flag. Maybe we set it not for VS only?

ihhub:
Hey @yannlabou , I found an information that it is possible to run ARM build on Travis-CI using QEMU ARM. Would you mind to try this? It would be a good opportunity to avoid ARM compilation errors in future.

I can create a separate issue for this if needed.

0x72D0:
I don't have my raspberry pi until monday, I'm gonna test it then.

yannlabou:
> Hey @yannlabou , I found an information that it is possible to run ARM build on Travis-CI using QEMU ARM. Would you mind to try this? It would be a good opportunity to avoid ARM compilation errors in future.
> 
> I can create a separate issue for this if needed.

Yes. We might need a separate issue. I'll see how much work is needed for that.

CMakeLists.txt
yannlabou: > I have a doubt that Visual Studio supports such compiler flag. Maybe we set it not for VS only?

Unknown flags are ignored during compilation though. I checked the logs and everything looks OK to me.

0x72D0:
ok I just tried compiling with cmake it and I got the following error:

image

here's what cmake output:

image
image

ihhub:
Nooooo.... :)
These errors are due to missing compilation flag with neon support.

0x72D0:
Yeah, I tried compiling the master branch with march=native and it's gave me these errors. While when I tried compiling it with mfpu=neon-vfpv4, the compilation works.

ihhub:
@0x72D0 try this way: instead of this code
```
if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)")
    add_compile_options(-mfpu=neon-vfpv4)
endif()
```
put this and check whether you see the message:
```
if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)")
    message(WARNING "This is ARM processor")
    add_compile_options(-mfpu=neon-vfpv4)
endif()
```

if you don't see please modify the code to like this only:
```
 add_compile_options(-mfpu=neon-vfpv4)
```

At least we would know where is the problem :)

0x72D0:
nope did't show up:
image
image

ihhub:
@0x72D0 please try just this line without any if condition:
```
add_compile_options(-mfpu=neon-vfpv4)
```

0x72D0:
ok when I put only:

```
add_compile_options(-mfpu=neon-vfpv4)
``` 

the program compile and the unit_test seems working.

yannlabou:
@0x72D0 Could you try the following?
```
message(STATUS "DEBUG ${CMAKE_SYSTEM_PROCESSOR}")
add_compile_options(-mfpu=neon-vfpv4)
```

0x72D0:
the architecture seem to be armv7l

image

0x72D0:
@yannlabou ok I just tried the version with the new regex and the compilation works fine now

ihhub:
@0x72D0 thanks a lot for testing!

@yannlabou do you have any more pending changes? I think it's good to commit now :)

yannlabou:
I pushed the regex fix 8 hours ago.

ihhub:
@yannlabou thanks a lot for the fix!

PR #467, opened 2019-05-19
Change imageException to std::exception in catch block in Raspicam example

### LINKED ISSUE #460: Change imageException to std::exception in catch block in Raspicam example
It is as simple as it states. We need to change line 101 in `examples/raspicam/main.cpp` to line:
```cpp
catch( const std::exception & ex)
```

This is needed to cover all possible standard exceptions as well.

theoniko:
Closes #460


ihhub:
@theoniko thanks a lot for your continious contribution to the project!

PR #473, opened 2019-05-24
Implement multithreaded version Split function

### LINKED ISSUE #364: Implement multithreaded version of Split() function (based on existing thread pool code)
We do not have function pool (multithreading based on thread pool) implementation of **Split()** function. For the person who is willing to add such code please take a look into an implementation of the function inside `image_function.cpp` and the structure of code for function pool inside `function_pool.cpp`. Please do not forget to add unit test for newly created function(s) inside `unit_test_image_function.cpp` (it's just one line code via macros call).

theoniko:
Closes #364


ihhub:
@theoniko thank you very much for your continuous support of the project!

Issue #469, opened 2019-05-21
Merge CUDA and Qt jobs for Linux in Travis-CI


ihhub:
Since we are going to expand unit tests and add more jobs (potential ARM support) we need to care about overall job execution time which is currently at around 1 hour. We could merge first 2 jobs which run on Linux machines for Cuda and Qt support. In such case we reduce execution time as we do not run unit tests twice.

ihhub:
Hi @yannlabou , would you mind to take a look at this issue after if you have time?

yannlabou:
This should be doable. I will do that. Not sure if this week but I will.

ihhub:
@yannlabou thanks a lot for taking this issue!

PR #479, opened 2019-05-30
Issue 469: Merging CUDA and Qt Travis jobs

yannlabou:
relates to #469

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,11 @@ addons:
 matrix:
   include:
     - os: linux
-      env: CUDA=8.0.61-1 CMAKE_OPTIONS="-DPENGUINV_BUILD_TEST=ON"
-      before_install: source script/travis/install_cuda.sh
-    - os: linux
-      env: QT_BASE=59 CMAKE_OPTIONS=""
-      before_install: source script/travis/install_qt5.sh
+      name: "CUDA8-Qt5.9"
+      env: CUDA=8.0.61-1 QT_BASE=59 CMAKE_OPTIONS="-DPENGUINV_BUILD_TEST=ON"
+      before_install:
+        - source script/travis/install_cuda.sh
+        - source script/travis/install_qt5.sh
     - os: linux
       env: STATIC_ANALYSIS=ON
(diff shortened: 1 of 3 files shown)


ihhub:
Is it correct to say that we fail Qt installation? I see that we have same error even previously.

yannlabou:
> Is it correct to say that we fail Qt installation? I see that we have same error even previously.



> Is it correct to say that we fail Qt installation? I see that we have same error even previously.

Yes! Thank you for spotting this bug. I'm trying to come up with a robust solution.

ihhub:
I think we need to fail the job if we cannot download Qt otherwise we don't build Qt example even. What do you think?

yannlabou:
> I think we need to fail the job if we cannot download Qt otherwise we don't build Qt example even. What do you think?

That's the purpose of [travis_terminate 1](https://github.com/ihhub/penguinV/pull/479/commits/d265e18f9ca25539c9065f99ffbd3f163f235062#diff-354f30a63fb0907d4ad57269548329e3R23)

ihhub:
Ah, okay.

yannlabou:
> Ah, okay.

travis_terminate didn't have the expected behaviour. I moved the ppa source from the install script to the .travis.yml. Works now.

.travis.yml
```diff
@@ -4,22 +4,26 @@ language: cpp
 
 addons:
   apt:
+    sources:
+      - sourceline: 'ppa:beineri/opt-qt596-xenial'
     packages:
+      - libgl-dev
```
ihhub: Could you please explain why do we need OpenGL support?

yannlabou: > Could you please explain why do we need OpenGL support?

Because Qt5.9 in Ubuntu Xenial seems to require GL as we can see [here](https://travis-ci.org/ihhub/penguinV/jobs/539579691#L540).

ihhub: Ah, okay. Thanks for explanation.

ihhub:
@yannlabou thank you very much for you excellent work!

PR #480, opened 2019-05-30
Optimise Dilate() function performance

### LINKED ISSUE #463: Optimise Dilate() function performance
We have **Dilate()** function in `src/image_function.cpp` under unnamed namespace. There are places where we could potentially speed up function's execution:
1. after reaching the condition `if( imageXEnd - imageX < dilateX )` or `if( height - rowId < dilationY )` we should stop loop as we reached end of line/column.
2. Initialize `const uint32_t rowSize = image.rowSize();` only once
3. Add check if `dilationX` and `dilationY` are equal to 0 then exit function

Additionally, as a part of future commit please remove one empty line before `if( dilationX > width / 2 )`.

theoniko:
Closes #463

### DIFF
--- a/src/image_function.cpp
+++ b/src/image_function.cpp
@@ -58,7 +58,7 @@ namespace
         Image_Function::ParameterValidation( image, x, y, width, height );
         Image_Function::VerifyGrayScaleImage( image );
 
-        if ( dilationX == 0u && dilationY == 0u ) 
+        if ( dilationX == 0u && dilationY == 0u )
             return;
 
         if( dilationX > width / 2 )
--- a/src/image_function.cpp
+++ b/src/image_function.cpp
@@ -94,9 +94,9 @@ namespace
                             startPos[pairCount] = imageX - dilateX;
 
                         if( imageXEnd - imageX < dilateX ) {
-                            endPos[pairCount] = imageXEnd;
-							break;
-						}
+			     endPos[pairCount] = imageXEnd;
+			     break;
+			 }
                         else
                             endPos[pairCount] = imageX + dilateX;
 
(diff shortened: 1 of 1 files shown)


src/image_function.cpp
```diff
@@ -58,19 +58,23 @@ namespace
         Image_Function::ParameterValidation( image, x, y, width, height );
         Image_Function::VerifyGrayScaleImage( image );
 
+	 if (dilationX == 0u && dilationY == 0u) {
```
ihhub: Please do not use `{` and `}` for single line code inside if block to make code look nicer :)

src/image_function.cpp
```diff
@@ -90,10 +93,13 @@ namespace
                         else
                             startPos[pairCount] = imageX - dilateX;
 
-                        if( imageXEnd - imageX < dilateX )
-                            endPos[pairCount] = imageXEnd;
-                        else
+                        if ( imageXEnd - imageX < dilateX ) {
+			     endPos[pairCount] = imageXEnd;
+			     break;
```
ihhub: We have to increase pairCount before break.

src/image_function.cpp
```diff
@@ -58,19 +58,22 @@ namespace
         Image_Function::ParameterValidation( image, x, y, width, height );
         Image_Function::VerifyGrayScaleImage( image );
 
-
+	 if ( dilationX == 0u && dilationY == 0u )
```
ihhub: @theoniko could you please check indentation of lines. Here we have 9 spaces instead of 8 :)

theoniko:
Hello @ihhub,

Thank you for your review. I will make the requested changes once i have free time.

ihhub:
@theoniko just a small note: please use 4 spaces instead of tabs according to your coding style :)

ihhub:
@theoniko thank you very much for your continuous support of the project!

PR #514, opened 2019-08-05
Fix makefile for unit tests (lpng and ljpeg flags) 

### LINKED ISSUE #509: Fix makefile for unit tests (lpng and ljpeg flags)
We have `test/unit_tests/Makefile` which contains `-lpng -ljpeg` flags for linking. If OS doesn't have such libraries we fail the compilation which we shouldn't do. The solution for this is to use an extra code like:
```
PNG_EXISTS = $(shell ldconfig -p | grep libpng)

ifneq ($(PNG_EXISTS),)
CXXFLAGS += -D PENGUINV_ENABLED_PNG_SUPPORT
LIBS += -lpng
endif
```

Just after settings all variables inside the file.

fullset:
Closes #509 

Hello, @ihhub 
Look, please.


ihhub:
@fullset could you please remove compiled application (1.58 MB) from pull request.

ihhub:
@fullset thank you very much for your continuous support of the project!

Issue #144, opened 2018-10-15
Makefile for CUDA performance tests


ihhub:
We do not have makefile for performance tests on CUDA but unit tests have. You could have a look at CUDA unit tests makefile to create a makefile for performance tests.

ihhub:
Hi @fullset would you mind to have a look into this task if you have free time for this and willingness to do such?

fullset:
hello, @ihhub
All right, I'm going to do this.

ihhub:
Thank you very much!

PR #524, opened 2019-08-21
Modify old third-party paths in makefile of CUDA unit tests

fullset:
@ihhub, look please.

When I was fixing  issue #144 I found out that makefile of cuda unit tests contains some old paths.
Maybe it would be better to fix it too?


ihhub:
@fullset thanks a lot for fixing this bug!

Issue #500, opened 2019-06-10
Make ParameterValidation() function as variadic template


ihhub:
We have multiple **ParameterValidation**() functions in `src/parameter_validation.h` file. We face a problem that we need to add 4th variation of the function to accept 4 images which looks like code duplication.
We want to avoid a restriction for number of parameters for this function as well as reduce duplication code so we need to make this function to accept variadic parameters. You can have a look at `VerifyGrayScaleImage` function as an example.

PR #525, opened 2019-08-25
Refactor parameter validation to variadic template

MeKot:
Completes #500

### DIFF
--- a/src/blob_detection.cpp
+++ b/src/blob_detection.cpp
@@ -15,7 +15,7 @@ namespace
         EDGE       = 3u,
         CONTOUR    = 4u
     };
-
+    
     double getLengthFromCountour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, PointBase2D< uint32_t > & startPoint,
                                  PointBase2D< uint32_t > & endPoint )
     {
@@ -304,7 +304,7 @@ namespace Blob_Detection
     const std::vector < BlobInfo > & BlobDetection::find( const PenguinV_Image::Image & image, uint32_t x, uint32_t y, uint32_t width,
                                                           uint32_t height, BlobParameters parameter, uint8_t threshold )
     {
-        Image_Function::ParameterValidation( width, height, image, x, y );
+        Image_Function::ParameterValidation( image, x, y, width, height );
         Image_Function::VerifyGrayScaleImage( image );
 
         parameter._verify();
--- a/src/blob_detection.cpp
+++ b/src/blob_detection.cpp
@@ -15,7 +15,7 @@ namespace
         EDGE       = 3u,
         CONTOUR    = 4u
     };
-    
+
     double getLengthFromCountour( const std::vector < uint32_t > & contourX, const std::vector < uint32_t > & contourY, PointBase2D< uint32_t > & startPoint,
                                  PointBase2D< uint32_t > & endPoint )
     {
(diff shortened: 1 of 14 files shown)


src/blob_detection.cpp
```diff
@@ -304,7 +304,7 @@ namespace Blob_Detection
     const std::vector < BlobInfo > & BlobDetection::find( const PenguinV_Image::Image & image, uint32_t x, uint32_t y, uint32_t width,
                                                           uint32_t height, BlobParameters parameter, uint8_t threshold )
     {
-        Image_Function::ParameterValidation( image, x, y, width, height );
+        Image_Function::ParameterValidation( width, height, image, x, y );
```
ihhub: Why did we change input variables for this function? Now we have to refactor whole library for this

src/parameter_validation.h
```diff
@@ -71,46 +71,26 @@ namespace Image_Function
...
-            startX2 + width > image2.width() || startY2 + height > image2.height() || startX2 + width < width || startY2 + height < height )
-            throw imageException( "Bad input parameters in image function" );
-    }
-
-    template <typename TImage>
-    void ParameterValidation( const TImage & image1, uint32_t startX1, uint32_t startY1,
-                              const TImage & image2, uint32_t startX2, uint32_t startY2,
-                              const TImage & image3, uint32_t startX3, uint32_t startY3,
-                              uint32_t width, uint32_t height )
+    template <typename TImage, typename... Args>
```
ihhub: Let me think please how to avoid such big change,,,

MeKot:
I can squash them if needed

src/blob_detection.cpp
MeKot: Yep, I know, but the only way to make images and their dimensions variadic, was to put them at the end of the function..

src/parameter_validation.h
MeKot: Sure! I am happy to change this if there is a better way!

ihhub: Okay. I got an idea: we have an extra function which returns `std::pair<uint32_t, uint32_t>` value. Let's call it `std::pair<uint32_t, uint32_t> ExtractRoiSize( uint32_t width, uint32_t height )`.

Now we need to declare template function:
```cpp
template <typename TImage, typename... Args>
std::pair<uint32_t, uint32_t> ExtractRoiSize( const TImage & image1, uint32_t startX1, uint32_t startY1, Args... args )
{
    return ExtractRoiSize( args );
}
```

So then in your original function `ParameterValidation` you could get width and height

ihhub: It looks little ugly but we don't need to change the rest of library.

MeKot: I will give it a go, thanks!

MeKot: @ihhub Could you please have another look?

src/parameter_validation.h
```diff
@@ -71,13 +72,22 @@ namespace Image_Function
...
     {
-        if( image1.empty() || image2.empty() || image3.empty() || !IsCorrectColorCount( image1 ) || !IsCorrectColorCount( image2 ) ||
-            !IsCorrectColorCount( image3 ) || image1.width() != image2.width() || image1.height() != image2.height() ||
-            image1.width() != image3.width() || image1.height() != image3.height() )
-            throw imageException( "Bad input parameters in image function" );
+        ParameterValidation( image1, image2 );
+        ParameterValidation( image2, args... );
+    }
+
+    inline std::pair<uint32_t, uint32_t> ExtractRoiSize( uint32_t width, uint32_t height )
```
ihhub: To avoid inline and cross file compilation let's change this function into template function something like this:
```cpp
template <typename _Type>
std::pair<_Type, _Type> ExtractRoiSize( _Type width, _Type height )
{
    return std::pair<_Type, _Type>( width, height );
}
```

MeKot: Sure! I was not sure if having such a generic function would be a good idea.. But that certainly makes life a lit easier

MeKot:
The build is failing because it cannot install `swig`

MeKot:
Sorry for the mess in one of the sources... Should this be merged now @ihhub, or are there any other outstanding issues?

ihhub:
@MeKot thank you very much for fixing this issue!

It took a lot of time for me for review as I'm handling two open source projects so second project was required some attention :)

PR #527, opened 2019-08-31
Add static cast to uint32_t in x11_ui.cpp

### LINKED ISSUE #494: Fix compilation warning in x11 function call
We follow the idea of zero warnings in source code. Unfortunately, it's not ease sometimes to do not make such warnings appear.
Right now we have one warning which should be fixed:
```bash
src/ui/x11/x11_ui.cpp: In member function ‘void UiWindowX11::_setupImage(const Image&)’:
src/ui/x11/x11_ui.cpp:107:68: warning: conversion to ‘unsigned int’ from ‘int’ may change the sign of the result [-Wsign-conversion]
                            0, _data.data(), _width, _height, 32, 0 );
```

You can have a look into this warning [here](https://travis-ci.org/ihhub/penguinV/jobs/543525142) (please ignore kissfft warnings for now).

MeKot:
Closes #494

### DIFF
--- a/src/ui/x11/x11_ui.cpp
+++ b/src/ui/x11/x11_ui.cpp
@@ -103,8 +103,8 @@ void UiWindowX11::_setupImage( const PenguinV_Image::Image & image )
     }
 
     const int defaultScreen = DefaultScreen( _uiDisplay );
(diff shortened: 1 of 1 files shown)


ihhub:
@MeKot thank you very much for fixing this issue!

Issue #508, opened 2019-07-03
Merge BMP and PNG unit test code


ihhub:
We have unit tests for Bitmap (`test/unit_tests/unit_test_bitmap.h` and `test/unit_tests/unit_test_bitmap.cpp` files) and for PNG (`test/unit_tests/unit_test_png.h` and `test/unit_tests/unit_test_png.cpp` files). We would like to merge then together as there is almost no difference in code. A new code should be located in `test/unit_tests/unit_test_file.h` and `test/unit_tests/unit_test_file.cpp` files.

PR #529, opened 2019-09-10
Do not run png unit tests if PNG support is disabled

ihhub:
relates to #508


ihhub:
Hi @fullset , here is a small fix which I forgot to make for your pull request :)

Issue #109, opened 2018-10-07
Add compilation for FFT CUDA code


ihhub:
Currently this code is not included in any tests. The compilation requires `cufft` library.

ihhub:
Hey, @fullset , would you like to try this?

fullset:
@ihhub I'm on vacation for now. I'm going to take a look at this on the next week.

fullset:
@ihhub, Could you explain a bit more detailed about what is required to do?

ihhub:
We need to add `cufft` library into compilation and add `src/cuda/cuda_fft.cuh`, `src/cuda/cuda_fft.cu`, `src/fft.cpp` and `src/fft.h` files into compilation list for CUDA unit tests.

PR #541, opened 2019-09-27
Add compilation for CUDA FFT

fullset:
WIP #109 
@ihhub, look please. I've added changes to Makefile and CMakefile but haven't done this for .vcxproj.


test/unit_tests/cuda/CMakeLists.txt
```diff
@@ -1,9 +1,17 @@
 set(MAIN_LIB_DIR ${CMAKE_SOURCE_DIR}/src)
+set(KISSFFT_DIR ${MAIN_LIB_DIR}/thirdparty/kissfft)
 set(UNIT_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR}/..)
 set(SRCS
     ${MAIN_LIB_DIR}/cuda/image_function_cuda.cu
     ${MAIN_LIB_DIR}/cuda/cuda_device.cu
     ${MAIN_LIB_DIR}/cuda/cuda_helper.cu
+    ${MAIN_LIB_DIR}/cuda/cuda_fft.cu
+    ${MAIN_LIB_DIR}/fft.cpp
```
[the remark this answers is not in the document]
fullset: I'm sorry but I added these files because you said about it in the issue explanation.

I've removed them already.

fullset:
@ihhub, look please for now.

I've tried to add cuda fft compilation for Windows using .vcxproj, but there some problems (according with appveyor log) which I can't to resolve because I'm not on Windows.

ihhub:
@fullset I'll take a look little later. I have Windows machine to verify your problem.

script/appveyor/cuda_install.cmd
```diff
@@ -7,7 +7,9 @@ cuda_8.0.44_windows.exe -s compiler_8.0 ^
                            cublas_dev_8.0 ^
                            cudart_8.0 ^
                            curand_8.0 ^
-                           curand_dev_8.0
+                           curand_dev_8.0 ^
+                           cufft_8.0 ^
+                           cufft_dev_8.0
```
ihhub: I'm just thinking do we really need normal and **dev** versions of the same package? What do you think?

fullset: Hmm, it looks like you're right. At least for cuda-cufft dev version depends on the normal version and they are both going to be installed. So we need only dev package.

ihhub:
@fullset AppVeyor build fails because `cufft` doesn't have implementation for Windows x86 (32-bit). One way is to put whole code for `src/cuda/cuda_fft.cuh` and `src/cuda/cuda_fft.cu` into **if !defined(_MSC_VER) && !defined(_WIN32)**

ihhub:
@fullset thanks a lot for your support to the project!

Issue #546, opened 2019-09-30
Make LookupTable() function for CUDA to support color images


ihhub:
We have an implementation of LookupTable() function for CUDA in `src/cuda/image_function_cuda.cu` file. This function doesn't support color image while an CPU implementation of the same function does (`src/image_function.cpp`). We need to add such support. Step to do:
1. Remove `Image_Function::VerifyGrayScaleImage( in, out );` line
2. Add these lines after verifying the size of table:
```cpp
const uint8_t colorCount = CommonColorCount( in, out );
width = width * colorCount;
```
3. Modify these lines:
```cpp
const uint8_t * inY  = in.data()  + startYIn  * rowSizeIn  + startXIn;
uint8_t       * outY = out.data() + startYOut * rowSizeOut + startXOut;
```

into
```cpp
const uint8_t * inY  = in.data()  + startYIn  * rowSizeIn  + startXIn * colorCount;
uint8_t       * outY = out.data() + startYOut * rowSizeOut + startXOut * colorCount;
```

PR #551, opened 2019-09-30
Added support in LookupTable() for CUDA to support color images (#546)

akhilgeothom:
Fixes issue ( #546 )
The variable width is changed according to color, but is not used anywhere else. Isn't this a concern?
The variable width could be passed by reference instead.

### DIFF
--- a/src/cuda/image_function_cuda.cu
+++ b/src/cuda/image_function_cuda.cu
@@ -912,12 +912,12 @@ namespace Image_Function_Cuda
             throw imageException( "Lookup table size is not equal to 256" );
 
         const uint8_t colorCount = CommonColorCount( in, out );
-        width = width * colorCount;  //this change is not reflected anywhere else
+        width = width * colorCount;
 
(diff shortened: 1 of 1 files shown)


src/cuda/image_function_cuda.cu
```diff
@@ -907,16 +907,18 @@ namespace Image_Function_Cuda
                       uint32_t width, uint32_t height, const std::vector < uint8_t > & table )
     {
         Image_Function::ParameterValidation( in, startXIn, startYIn, out, startXOut, startYOut, width, height );
-        Image_Function::VerifyGrayScaleImage( in, out );
 
         if ( table.size() != 256u )
             throw imageException( "Lookup table size is not equal to 256" );
 
+        const uint8_t colorCount = CommonColorCount( in, out );
+        width = width * colorCount;
```
ihhub: `width` is used at line 924 for **launchKernel2D** function/macro.

ihhub:
@akhilgeothom thank you very much for your contribution to the project!

Issue #538, opened 2019-09-26
Add CUDA implementation of SetPixel() function


ihhub:
We have an implementation of **SetPixel()** function for CPU which is stored in `src/image_function.h` and `src/image_function.cpp` files. This function has 2 overloads: one to setup one pixel only and another one is to setup an array of pixels. We would like to add the same function for **CUDA** code. We need to store this code in `src/cuda/image_function_cuda.cuh` and `src/cuda/image_function_cuda.cu` files.

gkarlos:
Hey, i was thinking of having a go on this. Are you looking for some sort of optimized kernel, or will a straightforward (somewhat naive) kernel do?

ihhub:
Naive kernel is enough for now.

gkarlos:
@ihhub quick question. Should SetPixel update the Image in place just like the CPU version?

ihhub:
@gkarlos yes, indeed. We could assume that image is very big so copying it is a bad idea.

PR #556, opened 2019-10-02
Cuda implementation of SetPixel() function

gkarlos:
As the title suggests, this PR solves #538 .  Unit tests included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@
 
 #build 
 /build/*
+
+#editors/IDEs
+/.vscode/*
--- a/.gitignore
+++ b/.gitignore
@@ -55,10 +55,3 @@
 *.fatbin
 *.cu.deps
 *.cu.cache
-
-
-#build 
-/build/*
-
-#editors/IDEs
-/.vscode/*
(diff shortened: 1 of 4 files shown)


gkarlos:
Looks like unit_test times out on the osx build? Any ideas why that might be?

ihhub:
@gkarlos I've restarted job. It might be time limit for pipeline job as 2 projects on my account receive many commits which trigger compilation on virtual machine. I'm working on reducing time for virtual machines.

.gitignore
```diff
@@ -55,10 +55,3 @@
 *.fatbin
 *.cu.deps
 *.cu.cache
-
-
-#build 
-/build/*
-
-#editors/IDEs
-/.vscode/*
```
gkarlos: I know these are unrelated to issue #538, but i think it'd be good if .gitignore included them. It's good not having to worry about staging build for instance. Are you interested in another PR that adds some common stuff to the .gitignore?

ihhub: @gkarlos please create a separate PR for this change if you think that this is valid :)

ihhub:
@gkarlos thank you very much for the contribution to the project!

Issue #548, opened 2019-09-30
Modify Histogram() function to support color images


ihhub:
We have an implementation of **Histogram()** function in `src/image_function.cpp` (without and with mask, there are 2 implementations). This function supports only gray-scale images for now. We want to add a support of color images as well. The output histogram for RGB image for example would be as follow red0, green0, blue0, red1, green1, blue1... In other words output array follows the same color channel structure an input image.
Steps to be done:
1. Remove `VerifyGrayScaleImage( image );` line
2. Change `histogram.resize( 256u );` into `histogram.resize( 256u * colorCount );`
3. Differentiate gray-scale image case (1 color channel) and multiple color channel so loops would be very efficient for every situation.

pierre87:
I would like to try this one :)

ihhub:
@pierre87 sure!

PR #557, opened 2019-10-03
Modify Histogram() function to support color images

pierre87:
Followed the instructions in #548

### DIFF
--- a/src/image_function.cpp
+++ b/src/image_function.cpp
@@ -858,23 +858,27 @@ namespace Image_Function
     void Histogram( const Image & image, uint32_t x, uint32_t y, uint32_t width, uint32_t height, std::vector < uint32_t > & histogram )
     {
         ParameterValidation( image, x, y, width, height );
-        VerifyGrayScaleImage( image );
         OptimiseRoi( width, height, image );
 
-        histogram.resize( 256u );
+        const uint8_t colorCount = image.colorCount();
+
+        histogram.resize( 256u * colorCount );
         std::fill( histogram.begin(), histogram.end(), 0u );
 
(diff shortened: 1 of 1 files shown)


src/image_function.cpp
```diff
@@ -858,23 +858,27 @@ namespace Image_Function
...
         for( ; imageY != imageYEnd; imageY += rowSize ) {
             const uint8_t * imageX    = imageY;
             const uint8_t * imageXEnd = imageX + width;
 
-            for( ; imageX != imageXEnd; ++imageX )
-                ++histogram[*imageX];
+            for( ; imageX != imageXEnd; imageX += colorCount ) {
+                uint8_t colorChannel = 0;
+                for( ; colorChannel < colorCount; colorChannel++ )
+                    ++histogram[*(imageX + colorChannel)];
```
ihhub: Your implementation will always fill 256 values of the array.

ihhub: Let's do this way:
```cpp
if ( colorCount  == 1 ) {
    // put original double loop here
} else {
    loop1
        loop2
            ++histogram[*(imageX + colorChannel) * colorCount + colorChannel];
}

pierre87: Oh yes you are right there is something wrong with `++histogram[...]`.
For the loops, sure thing I can fix that but can you please help me understand the benefit of splitting the cases? I thought it boils down to the same thing in case `colorCount` is 1.

ihhub: The benefit is performance only. Most of images with which people work are gray-scale aka 1 color channel image so it's pretty important to have highest performance.

ihhub:
@pierre87 thank you very much for the contribution to the project!

PR #561, opened 2019-10-05
Useful .gitignore additions

gkarlos:
Some useful .gitignore additions as discussed here https://github.com/ihhub/penguinV/pull/556#discussion_r331737013

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,17 @@
 *.fatbin
 *.cu.deps
 *.cu.cache
+
+# Build
+build/*
+
+# Editors/IDEs
+.vscode/*
+.history
+.eclipse/*
+./idea/*
+
+# OS generated
+.DS_store
+.Trashes
+._*


.gitignore
```diff
@@ -55,3 +55,17 @@
 *.fatbin
 *.cu.deps
 *.cu.cache
+
+# Build
+build/*
```
ihhub: Here we ignore specific directories. We should ignore specific files instead.

ihhub: There is a possibility when useful directory would have the same name as we would ignore it in git.

gkarlos: Are you talking about `build` specifically?

ihhub: I'm talking about every directory in your list. If directory is empty git won't show it in the list.

gkarlos: Sure yes. But we put it there for the cases that it isn't. I'm sorry i just think i don't get your point.

gkarlos: Also w.r.t your previous comment. Well i think its safe to say that almost everybody names their build folder `build`

ihhub: Could you please write a list of file extensions locate in this directory? For example such directory doesn't exist for Visual Studio (not Code) build. I'm not arguing, I'm just trying to find a correct explanation for this :)

gkarlos: Sure no worries. Me neither :) I was just really wondering why you think someone might use these directories for anything else within a project. In one laptop for instance i am using vscode. My `.vscode` folder has only a `settings.json` file. I am not sure if that's true for everyone (not my main editor so don't know that much), but if you do use vscode then such a folder is automatically created in the root of your project. Good to not track. The `.idea` folder from CLion probably has some `xml` files in. I guess its better to ignore these dirs only instead of all `.xml` of `.json` files. And i guess probably everybody names their build folder `build`. It would be weird to have a build folder that is used for something. The rest macOS generated files. I could try to give you a more concrete example, i.e. precisely which files are in these dirs, but i do not think this is necessarily true for everyone. It may depend on the setup.

ihhub:
@gkarlos thanks for you continuous contribution to the project!

PR #565, opened 2019-10-15
Change randomValue function to provide reaching of maximum value.

fullset:
Hey, @ihhub look, please.

You've said in the description that 1 on the way to solve this issue is 
> Change `rand() % maximum` to `rand() % (maximum + 1)`. After this we have to change all calls of this function.

So, I've changed the most of calls of `randomValue()` but some other calls look fine without changing.

### DIFF
--- a/test/unit_tests/cuda/unit_test_image_function_cuda.cpp
+++ b/test/unit_tests/cuda/unit_test_image_function_cuda.cpp
@@ -207,8 +207,8 @@ namespace image_function_cuda
             const std::vector < uint8_t > intensity = intensityArray( 2 );
             PenguinV_Image::Image input = uniformImage( intensity[0], 0, 0, reference );
 
-            const bool horizontalFlip = (randomValue<uint32_t>( 0, 2 ) == 0);
-            const bool verticalFlip = (randomValue<uint32_t>( 0, 2 ) == 0);
+            const bool horizontalFlip = (randomValue<uint32_t>( 0, 1 ) == 0);
+            const bool verticalFlip = (randomValue<uint32_t>( 0, 1 ) == 0);
             const uint32_t xCorrection = input.width() % 2;
             const uint32_t yCorrection = input.height() % 2;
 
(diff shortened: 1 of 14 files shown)


test/unit_tests/cuda/unit_test_image_function_cuda.cpp
```diff
@@ -207,8 +207,8 @@ namespace image_function_cuda
             const std::vector < uint8_t > intensity = intensityArray( 2 );
             PenguinV_Image::Image input = uniformImage( intensity[0], 0, 0, reference );
 
-            const bool horizontalFlip = (randomValue<uint32_t>( 0, 2 ) == 0);
-            const bool verticalFlip = (randomValue<uint32_t>( 0, 2 ) == 0);
+            const bool horizontalFlip = ( randomValue<uint32_t>( 0, 1 ) == 0 );
```
ihhub: Let's create **randomBoolValue()** function instead. No harm to do this but it makes code easier to read.

test/unit_tests/unit_test_image_function.cpp
```diff
@@ -1935,8 +1935,8 @@ namespace Function_Template
     {
         const std::vector < uint8_t > intensity = intensityArray( 2 );
         PenguinV_Image::Image image  = uniformImage( intensity[0] );
-        const uint32_t x = randomValue<uint32_t>( 0, image.width() );
-        const uint32_t y = randomValue<uint32_t>( 0, image.height() );
+        const uint32_t x = randomValue<uint32_t>( 0, image.width() - 1 );
```
ihhub: I'm not sure that this format is good as another developer would definitely forget to add `- 1`. What's your opinion?

test/unit_tests/cuda/unit_test_image_function_cuda.cpp
ihhub: For this function we could use `rand()` directly, no need to call `randomValue<>` function

PR #566, opened 2019-10-15
Replace penguinV_Image namespace to penguinV

### LINKED ISSUE #543: Rename PenguinV_Image namespace into penguinV
We would like to rename **PenguinV_Image** namespace which locates in `src/image_buffer.h`, `src/opencl/image_buffer_opencl.h` and `src/cuda/image_buffer_cuda.cuh` files into **penguinV**.

Please take note that in `src/penguinv/penguinv.h` file we have to remove `using namespace PenguinV_Image;` line after renaming.

fullset:
Closes #543

### DIFF
--- a/src/function_pool.cpp
+++ b/src/function_pool.cpp
@@ -741,7 +741,7 @@ namespace Function_Pool
     {
         Image_Function::ParameterValidation( in1, in2 );
 
-        return IsEqual( in1, 0, 0, in2, 0, 0, in1.width(), in1.height() );
+        return Function_Pool::IsEqual( in1, 0, 0, in2, 0, 0, in1.width(), in1.height() );
     }
 
     bool IsEqual( const Image & in1, uint32_t startX1, uint32_t startY1, const Image & in2, uint32_t startX2, uint32_t startY2,
@@ -858,7 +858,7 @@ namespace Function_Pool
     void Normalize( const Image & in, uint32_t startXIn, uint32_t startYIn, Image & out, uint32_t startXOut, uint32_t startYOut,
                     uint32_t width, uint32_t height )
     {
-        const std::vector<uint32_t> histogram = Histogram( in, startXIn, startYIn, width, height );
+        const std::vector<uint32_t> histogram = Function_Pool::Histogram( in, startXIn, startYIn, width, height );
         if ( histogram.size() != 256u )
             throw imageException( "Histogram size is not equal to 256" );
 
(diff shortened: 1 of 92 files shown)


ihhub:
Hi @fullset , thanks a lot for your effort but the original issue was to rename to `penguinv` without Capital P like in `src/penguinv/penguinv.h` file :(

fullset:
@ihhub, Oops.. My bad. But it's not a very problem to change it.

src/function_pool.cpp
```diff
@@ -741,7 +741,7 @@ namespace Function_Pool
     {
         Image_Function::ParameterValidation( in1, in2 );
 
-        return IsEqual( in1, 0, 0, in2, 0, 0, in1.width(), in1.height() );
+        return Function_Pool::IsEqual( in1, 0, 0, in2, 0, 0, in1.width(), in1.height() );
```
ihhub: Why do we need these changes?

fullset: That's because we have such problems (https://travis-ci.org/ihhub/penguinV/jobs/599022738):

`error: call of overloaded ‘IsEqual(const Image&, int, int, const Image&, int, int, uint32_t, uint32_t)’ is ambiguous`

Before renaming of `PenguinV_Image` namespace, there were 3 namespaces: `PenguinV_Image`,` Function_Pool` (which had functions like `IsEqual()`) and `penguinV` (which have the same functions too). Both of `Function_Pool` and `penguinV` included only definitions from `PenguinV_Image`.

After renaming all of the names from `PenguinV_Image` is present in `penguinV`. And `Function_Pool` is using all these names now. And there are several conflicts with several functions with equal prototypes.

ihhub: Ah, okay. I'll check at home why it's like this.

ihhub: It happens because of `#include "penguinv/penguinv.h"` file inclusion.

fullset:
Hi, @ihhub.
Do you know what the problem is present there?

I'm not sure that I know good enough how travis and appveyor build pull requests, but it looks like they take master branch with changes of PR. 

Since I don't have any problem with my local building and in my fork (https://travis-ci.com/fullset/penguinV/builds/138432488), maybe it is worth to rebase my branch from the last master and then try to find problems?

ihhub:
Hi @fullset could you please create another pull request to recreate these changes? I don't know what's going on so let's try this approach first. Please note about my changes which I made in this pull request.

fullset:
Hi, @ihhub.

No problem, I'll make another PR. Just give me a few days. I have no time right now.

Issue #86, opened 2018-04-23
Support of AVX-512


ihhub:
More modern CPUs can support AVX-512 so why don't use it? We don't have CPUs with AVX-512 support yet :(

ihhub:
@0x72D0 do you want any chance to obtain a hardware (CPU) for this? :)

0x72D0:
and also I don't know if I'm gonna have a lot of time with my university...

ihhub:
Ah, I see :( Education is very important thing in this world ;)

Anyway, if you would have a time and willingness then you might consider an opportunity to join this project as a collaborator :) You could read what it means on github help.

ihhub:
Hey @0x72D0 would you mind to come back this issue once you have enough free time? You don't need to have a processor with AVX-512 but a compiler to do such. Let's make a blind step to implement AVX-512 without proper hardware which could test in future.

0x72D0:
Hi @ihhub, I willl try this after #431

0x72D0:
Hi @ihhub I was thinking of doing this issue for the hacktoberfest :). What extension of AVX512 should we support? I think it's easier to target a specific cpu architecture since there's so many of them. I was thinking about F, CD, VL, DQ, BW since it's the avx512 extension used by the skylake architecture. Or we could target the icelake architecture with F, CD, VL, DQ, BW, IFMA, VBMI, VBMI2, VPOPCNTDQ, BITALG, VNNI, VPCLMULQDQ, GFNI, VAES

ihhub:
Hi @0x72D0 , the choice is yours. The most important is that compiler is able to support to it. I can help in this issue as much as I can!

0x72D0:
Ok this table found on [wikipedia](https://en.wikipedia.org/wiki/AVX-512):
![image](https://user-images.githubusercontent.com/16087328/66703245-79b04200-ecde-11e9-8b0f-0aa3be63b77b.png)

 show that all AVX-512 processor use F and CD. Since most of the 8-bit simd operation we use in this project appear in BW and BW always come with VL and DQ, I think we should taget F, CD, BW, VL and DQ.

0x72D0:
@ihhub I'm gonna implement three function only by PR to avoid a situation like #434

0x72D0:
I tried sde like proposed in #611 with the current master branch (81379fc43e35fd1e509cef7a3b72aea7bbffeb82) and I got the following failed unit_tests:

[...]

PR #570, opened 2019-10-20
BitwiseXor, ConvertTo16Bit, ConvertTo8Bit implementation for AVX512

0x72D0:
PR for #86 

I added yet again three new function for AVX512 support.

The performance_tests and unit_tests compile with march=skylake-avx512


ihhub:
Hi @0x72D0 thanks a lot for implementing this feature! Could you please post results from performance tests? I'd like to see them :)

0x72D0:
@ihhub sry, but I don't have a AVX512 compatible processor at home, so I only know that the code compile

Issue #320, opened 2018-12-31
Add few more functions to Line2DBase class


ihhub:
We have `Line2DBase` template class which locates in `src/math/math_base.h` file. We want to add few more functions into this class:
- `Point2DBase projection( const Point2DBase & point )` returns a point which is a projection on line of input point. Example: if we have a line which is parallel to X axis (goes from [0; 0] to [1; 0] point), an input point is [3; 4] then return point would be [3; 0]
- `Point2DBase opposite( const Point2DBase & point )` return an opposite/mirror to line point to input point. Example: if we have a line which is parallel to X axis (goes from [0; 0] to [1; 0] point), an input point is [3; 4] then return point would be [3; -4]

Names of functions could be changed if you have more appropriate names for them :)

PR #577, opened 2019-11-12
Add projection() and opposite() methods to LineBase2D class

fullset:
WIP #320 

Hi, @ihhub. Look, please. Do we need some unit tests for these methods?

### DIFF
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -204,6 +204,19 @@ public:
         return (distanceToLine < 0 ? -distanceToLine : distanceToLine);
     }
 
+    PointBase2D<_Type> projection( const PointBase2D<_Type> & point ) const
+    {
+        const _Type dot_product = _direction.x * (point.x - _position.x) + _direction.y * (point.y - _position.y);
+        const PointBase2D<_Type> offset( _direction.x * dot_product , _direction.y * dot_product );
+        return _position + offset;
+    }
+
+    PointBase2D<_Type> opposite( const PointBase2D<_Type> & point ) const
+    {
+        const PointBase2D<_Type> offset = point - projection(point);
+        return point - offset - offset;
+    }
+
     template <template <typename, typename...> class _container>
     static LineBase2D bestFittingLine( const _container< PointBase2D<_Type> > & points )
     {
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -206,15 +206,15 @@ public:
 
     PointBase2D<_Type> projection( const PointBase2D<_Type> & point ) const
     {
-        const _Type dot_product = _direction.x * (point.x - _position.x) + _direction.y * (point.y - _position.y);
-        const PointBase2D<_Type> offset( _direction.x * dot_product , _direction.y * dot_product );
+        const _Type dotProduct = _direction.x * (point.x - _position.x) + _direction.y * (point.y - _position.y);
+        const PointBase2D<_Type> offset( _direction.x * dotProduct, _direction.y * dotProduct );
         return _position + offset;
     }
 
     PointBase2D<_Type> opposite( const PointBase2D<_Type> & point ) const
     {
         const PointBase2D<_Type> offset = point - projection(point);
-        return point - offset - offset;
+        return point - 2 * offset;
     }
 
     template <template <typename, typename...> class _container>
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -46,8 +46,7 @@ struct PointBase2D
 
     bool operator == ( const PointBase2D & point ) const
     {
-        return ( pvmath::isEqual( x, point.x ) && pvmath::isEqual( y, point.y ) ) ||
-               ( pvmath::isEqual( x, point.x, point.x ) && pvmath::isEqual( y, point.y, point.y ) );
+        return pvmath::isEqual( x, point.x ) && pvmath::isEqual( y, point.y );
     }
 
     PointBase2D & operator += ( const PointBase2D & point )
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -73,9 +73,21 @@ struct PointBase2D
         return PointBase2D( x - point.x, y - point.y );
     }
 
+    template <typename T>
+    PointBase2D operator * ( const T & value ) const
+    {
+        return PointBase2D( value * x, value * y );
+    }
+
     _Type x;
     _Type y;
 };
+
+template <typename _Type, typename T>
+PointBase2D<_Type> operator * ( const T & value, const PointBase2D<_Type> & point )
+{
+    return PointBase2D<_Type>( value * point.x, value * point.y );
+}
 
 template <typename _Type>
 struct PointBase3D : public PointBase2D<_Type>
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -226,7 +226,7 @@ public:
     PointBase2D<_Type> opposite( const PointBase2D<_Type> & point ) const
     {
         const PointBase2D<_Type> offset = point - projection( point );
-        return point - 2 * offset;
+        return 2 * projection( point ) - point;
     }
 
     template <template <typename, typename...> class _container>
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -206,14 +206,14 @@ public:
 
     PointBase2D<_Type> projection( const PointBase2D<_Type> & point ) const
     {
-        const _Type dotProduct = _direction.x * (point.x - _position.x) + _direction.y * (point.y - _position.y);
+        const _Type dotProduct = _direction.x * ( point.x - _position.x ) + _direction.y * ( point.y - _position.y );
         const PointBase2D<_Type> offset( _direction.x * dotProduct, _direction.y * dotProduct );
         return _position + offset;
     }
 
     PointBase2D<_Type> opposite( const PointBase2D<_Type> & point ) const
     {
-        const PointBase2D<_Type> offset = point - projection(point);
+        const PointBase2D<_Type> offset = point - projection( point );
         return point - 2 * offset;
     }
 
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -51,7 +51,7 @@ struct PointBase2D
 
     bool operator != ( const PointBase2D & point ) const
     {
-        return !(*this == point);
+        return !( *this == point );
     }
 
     PointBase2D & operator += ( const PointBase2D & point )
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -46,7 +46,8 @@ struct PointBase2D
 
     bool operator == ( const PointBase2D & point ) const
     {
-        return pvmath::isEqual( x, point.x ) && pvmath::isEqual( y, point.y );
+        return ( pvmath::isEqual( x, point.x ) && pvmath::isEqual( y, point.y ) ) ||
+               ( pvmath::isEqual( x, point.x, point.x ) && pvmath::isEqual( y, point.y, point.y ) );
     }
 
     PointBase2D & operator += ( const PointBase2D & point )
--- a/src/math/math_base.h
+++ b/src/math/math_base.h
@@ -49,6 +49,11 @@ struct PointBase2D
         return pvmath::isEqual( x, point.x ) && pvmath::isEqual( y, point.y );
     }
 
+    bool operator != ( const PointBase2D & point ) const
+    {
+        return !(*this == point);
+    }
+
     PointBase2D & operator += ( const PointBase2D & point )
     {
         x += point.x;
@@ -225,7 +230,6 @@ public:
 
     PointBase2D<_Type> opposite( const PointBase2D<_Type> & point ) const
     {
-        const PointBase2D<_Type> offset = point - projection( point );
         return 2 * projection( point ) - point;
     }
 
--- a/test/unit_tests/unit_test_math.cpp
+++ b/test/unit_tests/unit_test_math.cpp
@@ -118,17 +118,20 @@ namespace pvmath
     template <typename _Type>
     bool pointProjection()
     {
-        PointBase2D<_Type> point0( 0, 0 );
-        PointBase2D<_Type> pointX( 1, 0 );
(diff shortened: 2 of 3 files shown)


ihhub:
Hi @fullset , yes, we need some unit tests to make sure that it works. You could assume in the tests that line is vertical or horizontal so it's easy to calculate where projection and opposite points would be.

fullset:
Ok. I've added.

test/unit_tests/unit_test_math.cpp
```diff
@@ -114,6 +114,56 @@ namespace pvmath
         }
         return true;
     }
+
+    template <typename _Type>
+    bool pointProjection()
+    {
+        PointBase2D<_Type> point1( 0, 0 );
```
ihhub: Hi @fullset is it possible to create a random based values in a loop instead of hardcoded values?

fullset: I guess it's possible. I just need to find out a good converting way from coordinates to projections.

ihhub: It's easy: you could assume that your line is vertical or horizontal and then third point is somehow locates at orthogonal line (vertical or horizontal) or at the same line.

fullset: Oh, you're right. I was thinking about a universal formula.

fullset: I guess I've done it.

test/unit_tests/unit_test_math.cpp
```diff
@@ -114,6 +114,46 @@ namespace pvmath
         }
         return true;
     }
+
+    template <typename _Type>
+    bool pointProjection()
+    {
+        PointBase2D<_Type> point0( 0, 0 );
```
ihhub: What if we have random data for these 3 points as well:
```cpp
PointBase2D<_Type> point0( randomValue(), randomValue() );
PointBase2D<_Type> point1( point0.x, randomValue() );
PointBase2D<_Type> point2( randomValue(), point.y );
```

Moreover we could something like this:
```cpp
PointBase2D<_Type> pointBase( randomValue(), randomValue() );
const LineBase2D<_Type> lineX( pointBase, PointBase2D<_Type>( pointBase.x, randomValue() ) );
const LineBase2D<_Type> lineY( pointBase, PointBase2D<_Type>( randomValue(), pointBase.y ) );
```

Looks better?

fullset: Looks interesting!

fullset: Hmm, I've got an unexpected problem with rounding of float values. Tests with doubles are OK, but tests with floats don't pass. I'm going to solve this problem, but it might take a while.

ihhub: Could you specify details of your problem? I faced the same issue some time ago?

fullset: Ok. I made lines to be based on random values:

```
const PointBase2D<_Type> pointBase( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ) );
const PointBase2D<_Type> pointBaseX( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), pointBase.y );
const PointBase2D<_Type> pointBaseY( pointBase.x, Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ) );

const LineBase2D<_Type> lineX( pointBase, pointBaseX );
const LineBase2D<_Type> lineY( pointBase, pointBaseY );
```

After starting unit_tests I get this:
```
341 of 342 tests passed.
List of failed tests: 
math::Line2d point opposition (float)
```

I calculated the difference between real values of opposition and expected values:

```
std::cout << lineX.opposite( testPoint ).x - resultPointX.x << " " << lineX.opposite( testPoint ).y - resultPointX.y << std::endl;
std::cout << lineY.opposite( testPoint ).x - resultPointY.x << " " << lineY.opposite( testPoint ).y - resultPointY.y << std::endl;
```

So, in `math::Line2d point opposition (double)` I get:
```
0 0
0 0
0 0
0 0
...
0 0
0 0
```

But in `math::Line2d point opposition (double)` test the result a bit different:
```
[342/342] math::Line2d point opposition (float)... 
0 0
0 0


0 0
0 0


0 0
0 3.051757812e-05

FAIL
```
Or even:
```
[342/342] math::Line2d point opposition (float)... 
0 0
0 0


0 0
0 0


0 0
0 0


0.0001220703125 0
0 0

FAIL
```

ihhub: This is definitely due to limitations of float format.

src/math/math_base.h
```diff
@@ -204,6 +216,19 @@ class LineBase2D
...
+    {
+        const _Type dotProduct = _direction.x * ( point.x - _position.x ) + _direction.y * ( point.y - _position.y );
+        const PointBase2D<_Type> offset( _direction.x * dotProduct, _direction.y * dotProduct );
+        return _position + offset;
+    }
+
+    PointBase2D<_Type> opposite( const PointBase2D<_Type> & point ) const
+    {
+        const PointBase2D<_Type> offset = point - projection( point );
+        return point - 2 * offset;
```
ihhub: Let's try this way:
```cpp
return 2 * projection( point ) - point;
```

which comes from this:
```
point - 2 * offset
point - 2 * ( point - projection( point ) )
point - 2 * point + 2 * projection( point )
-point + 2 * projection( point )
2 * projection( point ) - point
```

Such approach will improve an accuracy of calculations.

test/unit_tests/unit_test_math.cpp
ihhub: I added another comment above with suggested accuracy improvement. Could you please rerun tests after?

src/math/math_base.h
fullset: You know, I tried this way. The problems still present.

I have another suggestion how to solve this problem but you could don't like it because it changes the `pwmath::isEqual()` method. Look at my last commit, please.  [c5fdd1ad](https://github.com/ihhub/penguinV/pull/577/commits/c5fdd1adcccdadfba1405beae5d1d197009c02ed)

fullset: That might be because float rounds up to 7 digits. When we have little numeral (up to 100), 7 digits are 99.99999. And it's inside the accuracy of 1e-5. But if we have the point(-100, -999) and want to find opposition of line ((0, 999)(1, 999)), we'll get the point(-100, 1998). And 7 digits will be 1998.000 which are not enough to provide accuracy of 1e-5.

ihhub: No, it's incorrect. Float point variables are under IEEE 754. We loose an accuracy because of non-ideal computations. Please try with my changes and see the accuracy after.

test/unit_tests/unit_test_math.cpp
```diff
@@ -114,6 +114,50 @@ namespace pvmath
...
+
+    template <typename _Type>
+    bool pointProjection()
+    {
+        for ( uint32_t i = 0; i < Unit_Test::runCount(); ++i ) {
+            const PointBase2D<_Type> testPoint( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ) );
+
+            const PointBase2D<_Type> pointBase( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ) );
+            const PointBase2D<_Type> pointBaseX( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), pointBase.y );
+            const PointBase2D<_Type> pointBaseY( pointBase.x, Unit_Test::randomFloatValue<_Type>( -330, 330, 0.01f ) );
```
ihhub: Why do we use -330 to 330 range instead of -1000 to 1000?

src/math/math_base.h
ihhub: Could you please revert your changes for now and show me results after my suggested change. The problem lies in number of computations we do for opposite point. Less computations - more precise results.

ihhub: Let me find some info which would simplify the solution...

test/unit_tests/unit_test_math.cpp
fullset: It's just my mistake.

test/unit_tests/unit_test_math.cpp
```diff
@@ -114,6 +114,50 @@ namespace pvmath
...
+
+            const LineBase2D<_Type> lineX( pointBase, pointBaseX );
+            const LineBase2D<_Type> lineY( pointBase, pointBaseY );
+
+            const PointBase2D<_Type> resultPointX( testPoint.x, pointBase.y );
+            const PointBase2D<_Type> resultPointY( pointBase.x, testPoint.y );
+
+            if ( !( lineX.projection( testPoint ) == resultPointX ) || !( lineY.projection( testPoint ) == resultPointY ) )
```
ihhub: `if ( ( lineX.projection( testPoint ) != resultPointX ) || ( lineY.projection( testPoint ) != resultPointY ) )` is simpler :)

test/unit_tests/unit_test_math.cpp
```diff
@@ -114,6 +114,50 @@ namespace pvmath
...
+            const PointBase2D<_Type> pointBaseX( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), pointBase.y );
+            const PointBase2D<_Type> pointBaseY( pointBase.x, Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ) );
+
+            const LineBase2D<_Type> lineX( pointBase, pointBaseX );
+            const LineBase2D<_Type> lineY( pointBase, pointBaseY );
+
+            const PointBase2D<_Type> resultPointX( testPoint.x, pointBase.y * 2 - testPoint.y );
+            const PointBase2D<_Type> resultPointY( pointBase.x * 2 - testPoint.x, testPoint.y );
+
+            if ( !( lineX.opposite( testPoint ) == resultPointX ) || !( lineY.opposite( testPoint ) == resultPointY ) )
```
ihhub: `if ( ( lineX.opposite( testPoint ) != resultPointX ) || ( lineY.opposite( testPoint ) != resultPointY ) )`

test/unit_tests/unit_test_math.cpp
fullset: Of course.
I just was a bit lazy to implement `bool operator != ( const PointBase2D & point ) const `.

test/unit_tests/unit_test_math.cpp
```diff
@@ -114,6 +114,50 @@ namespace pvmath
...
+            const PointBase2D<_Type> pointBaseX( Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ), pointBase.y );
+            const PointBase2D<_Type> pointBaseY( pointBase.x, Unit_Test::randomFloatValue<_Type>( -1000, 1000, 0.01f ) );
+
+            const LineBase2D<_Type> lineX( pointBase, pointBaseX );
+            const LineBase2D<_Type> lineY( pointBase, pointBaseY );
+
+            const PointBase2D<_Type> resultPointX( testPoint.x, pointBase.y );
+            const PointBase2D<_Type> resultPointY( pointBase.x, testPoint.y );
+
+            if ( lineX.projection( testPoint ) != resultPointX || lineY.projection( testPoint ) != resultPointY )
```
ihhub: To fix accuracy issue we could do something like this:
`if ( pvmath::isEqual( lineX.projection( testPoint ), resultPointX, 10 ) || pvmath::isEqual( lineY.projection( testPoint ), resultPointY, 10 ) )`
by increasing accuracy threshold value by 10. Right now threshold value is 1e-5. It would be 1e-4.

fullset: This could help. But why we could use 10 as the multiplier? It looks like a magic number.

For example, in `pointOpposition()` unit test, I get calculation error `0.0001220703125` sometimes. And to fix the accuracy issue in such way, I need to use 15 as the multiplier.
And if we had random values from -100000 to 100000, we would have to increase the threshold value even more.

ihhub: We both understand that the problem is not with the code but with IEEE754 format. One solution is to change the range of random values. Another solution is comparison with ground truth value. We could go with first solution for example.

fullset: Hi, @ihhub.

Sorry for the late answer. 
What if the user of library will use `opposite()` method with big numbers (more than 10000)? He could find the same problem. It should be obvious for him to use `isEqual()` method instead of ==

ihhub: We cannot enhance standard float based computations so it's user's responsibility to used `float` instead of `double` if there is a need to do such.
We use epsilon value for comparison so I think end user could change this value instead.

ihhub: To have a fair results we need to use a square root of epsilon fo sigma which is `3e-5` (or 0.00003). Let me run your code on my machine and see values...

ihhub: For 90 degree angle we have `[4.37113883e-08, 1]` values of directories instead of `[0, 1]`. Value `4.37113883e-08` is standard error for float variable. The difference between possible generated points is 200. We multiply 2000 by 2 and by `4.37113883e-08` which gives us `0.0001748455532`. This explains why we cannot pass tests.

ihhub: So let's write a helper function here to compare two points with epsilon equal to 0.0002 with some explanation why we do this.

fullset: Do you mean a helper function inside `unit_test_math.cpp`?

ihhub: Yes.

ihhub: I added :)

ihhub:
Hi @fullset thank you very much for your continuous contribution to the project!

PR #578, opened 2019-11-14
Add Invert, Maximum, Minimum functions for AVX512

0x72D0:
PR for #86

I added yet again three new function for AVX512 support.
The performance_tests and unit_tests compile with march=skylake-avx512.

I can't test if the unit and performance tests work because I don't have a AVX512 processor in my possession.

I also fixes some missing code of my previous commit.

### DIFF
--- a/src/image_function_simd.cpp
+++ b/src/image_function_simd.cpp
@@ -2532,7 +2532,7 @@ if ( simdType == neon_function ) { \
         const uint32_t totalSimdWidth = simdWidth * simdSize;
         const uint32_t nonSimdWidth = width - totalSimdWidth;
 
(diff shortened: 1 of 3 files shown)


src/image_function_simd.cpp
```diff
@@ -2437,7 +2532,7 @@ if ( simdType == neon_function ) { \
         const uint32_t totalSimdWidth = simdWidth * simdSize;
         const uint32_t nonSimdWidth = width - totalSimdWidth;
 
-        AVX512BW_CODE( avx::BitwiseOr( rowSizeIn1, rowSizeIn2, rowSizeOut, in1Y, in2Y, outY, outYEnd, simdWidth, totalSimdWidth, nonSimdWidth ); )
```
ihhub: Good catch! I have an idea how to avoid such mistakes in future.

ihhub:
@0x72D0 thanks a lot for your efforts towards completing this huge task!

PR #580, opened 2019-11-21
Add missing UI functions for X11 Window class

### LINKED ISSUE #459: Add missing UI functions for X11 Window class
We have an optional UI Window classes for image display and drawings. All of them locate in `src/ui` directory. Currently we support Qt, Win32 API and X11 frameworks. **X11** code locates in `src/ui/x11/x11_ui.h` and `src/ui/x11/x11_ui.cpp` files. **UiWindowX11** class does not have an implementation of following functions:
```cpp
virtual void drawLine( const Point2d & start, const Point2d & end, const PaintColor & color );
virtual void drawEllipse( const Point2d & center, double xRadius, double yRadius, const PaintColor & color );
virtual void drawRectangle( const Point2d & topLeftCorner, double width, double height, const PaintColor & color );
```

We want to add them. You could have a look at Win32 or Qt implementation for reference of code implementation.

fullset:
Closes #459 

@ihhub 
I've added 3 methods to `UiWindowX11` class. I've made them similar to `drawPoint()` method. 
Also, I've slightly changed `drawPoint()` method to draw the cross(x).

Review please as you'll have a time.


src/ui/x11/x11_ui.cpp
```diff
@@ -52,8 +52,42 @@ void UiWindowX11::_display()
             for ( size_t i = 0u; i < _point.size(); ++i ) {
                 const Point2d & point = _point[i].first;
                 XSetForeground( _uiDisplay, defaultGC, _point[i].second );
-                XDrawLine( _uiDisplay, _window, defaultGC, static_cast<int>(point.x - 1), static_cast<int>(point.y - 1),
-                           static_cast<int>(point.x + 1), static_cast<int>(point.y + 1) );
+                XDrawLine( _uiDisplay, _window, defaultGC, static_cast<int>( point.x - 1 ), static_cast<int>( point.y - 1 ), static_cast<int>( point.x + 1 ),
```
ihhub: Hi @fullset could you please revert these changes as they are not relevant to this issue?

src/ui/x11/x11_ui.cpp
```diff
@@ -52,8 +52,42 @@ void UiWindowX11::_display()
...
+                           static_cast<int>( point.y + 1 ) );
+                XDrawLine( _uiDisplay, _window, defaultGC, static_cast<int>( point.x - 1 ), static_cast<int>( point.y + 1 ), static_cast<int>( point.x + 1 ),
+                           static_cast<int>( point.y - 1 ) );
+            }
+
+            for ( size_t i = 0u; i < _lines.size(); ++i ) {
+                const Point2d & start = std::get<0>( _lines[i] );
```
ihhub: I'm thinking to combine the code with `/src/ui/win/win_ui.h` file where we use structures to store data instead of tuples. Reading such code is very difficult so we might need to use the same idea in this class.

ihhub: We could move structures from `win_ui.h` file into `src/ui/ui.h`. The only thing missing is we need an integer value instead of 3 **uint8_t** values for color.

ihhub: We could do the same as it's done for Win32: just calculate RGB value during drawing (add a function at the top of the file to do such).

fullset: Ok, it sounds good. I'll do this.

ihhub: Great!

ihhub:
Hi @fullset let's finish this pull request :)

src/ui/x11/x11_ui.h
```diff
@@ -26,7 +30,64 @@ class UiWindowX11 : public UiWindow
     uint32_t _width;
     uint32_t _height;
 
-    std::vector< std::pair<Point2d, uint32_t> > _point;
+    struct PointToDraw
```
ihhub: Hi @fullset could we move these identical structures in a common location as src/ui/ui.h file to avoid code duplication?

src/ui/x11/x11_ui.cpp
```diff
@@ -109,7 +141,25 @@ void UiWindowX11::_setupImage( const PenguinV_Image::Image & image )
...
+
+void UiWindowX11::drawLine( const Point2d & start, const Point2d & end, const PaintColor & color )
+{
+    _lines.push_back( LineToDraw( start, end, ( color.red << 16 ) + ( color.green << 8 ) + color.blue ) );
+}
+
+void UiWindowX11::drawEllipse( const Point2d & center, double xRadius, double yRadius, const PaintColor & color )
+{
+    // XDrawArc needs x and y coordinates of the upper-left corner of the bounding rectangle but not the center of the ellipse.
+    Point2d position( center.x - xRadius, center.y - yRadius );
```
ihhub: Make this variable const.

fullset: done

src/ui/x11/x11_ui.h
fullset: As I see, there are 2 structures `PointToDraw` and `LineToDraw` which are almost identical in `x11` and `win_ui`. And there are no problem to combine them.

But `EllipseToDraw` structures in `x11` and `win_ui` have different semantics. In `win_ui` we specify top, left, right and bottom coordinates but it `x11` we specify top, left coordinates and width and height. I could make common constructor of 4 doubles but I'm not sure that it's right decision. What do you think?

ihhub: Let's stick to one design which should have top-left coordinate and width with height. For Windows during drawing we could easily find right and bottom values by adding width and height.

fullset: Hi, @ihhub.

It looks like I've broken windows building by my last changes but I can't find the reason because I haven't windows PC.

ihhub: Let me check...

ihhub: Could you put structures back in ui/ui.h but don't declare these variables:
```cpp
    std::vector<PointToDraw> _point;
    std::vector<LineToDraw> _lines;
    std::vector<EllipseToDraw> _ellipses;
```
?

Issue #582, opened 2019-12-04
Fix compilation with AVX 512 support


ihhub:
We have compilation error with `_mm512_set_epi8` function which you could find in [this compilation log](https://travis-ci.org/ihhub/penguinV/jobs/619688076). We need urgently fix it.

ihhub:
Hi @0x72D0 most likely problem lies in different set of avx512 functions. What we detect in cpuid might not support this function.

It's strange that compilation fails just recently.

ihhub:
The issue is fixed.

PR #583, opened 2019-12-04
add more condition to enable avx512 compilation

0x72D0:
here's a hotfix for #582.

It should solve the compilation problem only. We still need to add the flag identification for all the other avx512 module in cpu_id_unix.h and cpu_id_windows.h. Maybe also renaming PENGUINV_AVX512BW_SET define for PENGUINV_AVX512_SKYLAKE_SET since it's more accurate

### DIFF
--- a/src/penguinv/cpu_identification.h
+++ b/src/penguinv/cpu_identification.h
@@ -37,7 +37,15 @@
         #endif
 
         #ifdef __AVX512BW__
-            #define PENGUINV_AVX512BW_SET
+            #ifdef __AVX512CD__
+                #ifdef __AVX512DQ__
+                    #ifdef __AVX512F__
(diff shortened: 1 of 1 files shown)


0x72D0:
@ihhub this error is quite weird, can you take a look into it: https://travis-ci.org/ihhub/penguinV/jobs/620468777?utm_medium=notification&utm_source=github_status

I don't think it's avx512 fault there because it didn't run a avx512 function...

ihhub:
Hi @0x72D0 I'm looking into this...

ihhub:
Hi @0x72D0 please use `#if defined(AAA) && defined(BBB)...` command instead of nested structure.

ihhub:
Hi @0x72D0 thanks for the fix! I'll try to handle the rest.

PR #588, opened 2019-12-07
Overwriting ctest MemoryCheckCommandOptions

yannlabou:
Set Valgrind option "--show-reachable=no"
See http://valgrind.org/docs/manual/mc-manual.html


ihhub:
@yannlabou thanks a lot for the fix. I hope it would stop our problem with dynamic analysis

PR #589, opened 2019-12-07
AVX-512 code refactoring

0x72D0:
PR for #582 #86.

Since we are now not only detecting AVX512BW, this PR check if AVX512CD, AVX512DQ, AVX512F and AVX512VL is available by the CPU. Also, this PR rename AVX512BW for AVX512SKL.

### DIFF
--- a/src/penguinv/cpu_identification.h
+++ b/src/penguinv/cpu_identification.h
@@ -16,6 +16,8 @@
         #ifdef _MSC_VER
             #if _MSC_VER >= 1700
                 #define PENGUINV_AVX_SET
+            #elif _MSC_VER >= 1911
+                #define PENGUINV_AVX_SET
             #endif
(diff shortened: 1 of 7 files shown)


src/penguinv/cpu_identification.h
```diff
@@ -17,6 +17,9 @@
             #if _MSC_VER >= 1700
                 #define PENGUINV_AVX_SET
             #endif
+            #if _MSC_VER >= 1911
```
ihhub: Hi @0x72D0 I modified the condition here as logically it should be this way.

0x72D0: Yeah I agree it's better this way :)

ihhub:
Hi @0x72D0 thanks a lot for your valuable contribution to the project!

PR #592, opened 2019-12-24
AVX-512 implementation for ProjectionProfile, Sum and Subtract

0x72D0:
PR for #86 
I've added the support for avx512 to ProjectionProfile, Sum and Substract. It compile with -march=skylake-avx512.


ihhub:
@0x72D0 thanks you very much for your enormous contribution to the project!

PR #620, opened 2020-01-11
Treat warnings as errors

ihhub:
We don't want to pollute the project by warnings


ihhub:
It's bad idea as new version of compilers might introduce more warnings.

PR #636, opened 2020-01-18
Add missing code for SIMD functions

ihhub:
relates to #592


ihhub:
Hi @0x72D0 I think this might be a root of problem.

PR #673, opened 2020-10-04
Add script to output content of log files within a folder

### LINKED ISSUE #253: Add a script to output contents of log files within a folder
In Travis-CI script (`.travis`) for dynamic analysis we output valgrind results into log files. It's done by MemChecker itself. Currently we output results of log file with fixed name:
```bash
- cat "${TRAVIS_BUILD_DIR}/build/Testing/Temporary/MemoryChecker.2.log" > out && cat out
```
The main attention is on number 2 which is actually hardcoded. It's not a perfect solution.
To avoid this we need to create a script which would find all files with **.log** extension and output all contents of these files.

manekenpix:
Closes #253 

This adds a script to display the content of all log files in a folder.


ihhub:
@manekenpix thank you very much for fixing this issue!

PR #675, opened 2020-10-14
imageException renamed into penguinVException

### LINKED ISSUE #615: Rename imageException into penguinVException
We have a file `src/image_exception.h` which contains **imageException** class. This class is used for throwing exceptions. We would like to rename this class into **penguinVException** to have more accurate code structure. Please note that this exception is used in many places so we need to replace it everywhere.

kgorecki:
Resolves #615


ihhub:
Hi @kgorecki , great stuff! Just two things we need to do here:
- rename penguin_v_exception.h into penguinv_exception.h
- include penguinv_exception.h into the pull request and remove the old header.

ihhub:
@kgorecki thank you so much for fixing this issue!

PR #676, opened 2020-10-15
added CMake to .gitignore

kgorecki:
There's no issue for that (I didn't want to create any since it's minor). I recently noticed that there are no CMake related entries in .gitignore so I have added them.


ihhub:
Hi @kgorecki cmake files could be changed in the future so it's not rational to include them into gitignore list.

PR #678, opened 2021-10-01
Closes #644: Optimise Bitmap image saving function

### LINKED ISSUE #644: Optimise Bitmap image saving function
We have Bitmap image saving function `void Save( const std::string & path, const penguinV::Image & image, uint32_t startX, uint32_t startY, uint32_t width, uint32_t height )` which locates in ***src/file/bmp_image.h*** and ***src/file/bmp_image.cpp*** files.

During file saving we purposely copy a line of image to temporary array and then write the array into file. The reason behind this is that Bitmap image must have number of bytes to be as a multiplier of 4. For example, if image is 10 bytes/pixel it means that we need to write an additional 2 pixels of value 0.

We want to avoid extra copying to temporary array and directly write data from Image object and add needed amount of bytes into the file.

manekenpix:
Closes #644 

These changes optimize the [image saving function](https://github.com/ihhub/penguinV/blob/fa0b34e790ce8f0e149c8b4f1eb2a4819fc28226/src/file/bmp_image.cpp#L523) in **src/file/bmp_image.cpp**.

During the saving process, instead of copying a line of the pass image to a temporary array, along with the extra bytes needed for proper alignment, and then writing it into the output file; it writes the line from the image and the extra bytes for proper alignment directly into the file.

### DIFF
--- a/src/file/bmp_image.cpp
+++ b/src/file/bmp_image.cpp
@@ -575,14 +575,15 @@ namespace Bitmap_Operation
 
         const uint8_t * imageY = image.data() + (startY + height - 1) * rowSize + startX * colorCount;
 
-        std::vector < uint8_t > temp( lineLength, 0 );
-
         const size_t imageLineSize = sizeof( uint8_t ) * width * colorCount;
 
-        for ( uint32_t rowId = 0; rowId < height; ++rowId, imageY -= rowSize ) {
-            memcpy( temp.data(), imageY, imageLineSize );
+        const std::streamsize alignmentDiff = static_cast<std::streamsize>(lineLength - imageLineSize);
+        const uint32_t zero = 0;
 
-            file.write( reinterpret_cast<const char *>( temp.data() ), static_cast<std::streamsize>( lineLength ) );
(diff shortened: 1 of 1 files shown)


src/file/bmp_image.cpp
```diff
@@ -575,14 +575,15 @@ namespace Bitmap_Operation
 
         const uint8_t * imageY = image.data() + (startY + height - 1) * rowSize + startX * colorCount;
 
-        std::vector < uint8_t > temp( lineLength, 0 );
-
         const size_t imageLineSize = sizeof( uint8_t ) * width * colorCount;
 
+        uint8_t alignmentDiff = static_cast<uint8_t>(lineLength - imageLineSize);
```
ihhub, line 580: Please make this variable const and also it should be std::streamSize type to avoid extra type conversation.

ihhub (changes requested):

src/file/bmp_image.cpp
manekenpix: I made the requested changes

ihhub:
@manekenpix , thank you so much for your contribution!

PR #689, opened 2022-10-18
Information about installing libpng and libjpg is OS-dependent

### LINKED ISSUE #604: For MacOS missing libjpeg/libpng message should be different in makefile
Current message in **makefile** about missing libjpeg/libpng file is
```
Note! libjpeg-dev is missing from your computer,
so .jpeg images are not supported.

You can install libjpeg-dev by:
sudo apt install libjpeg-dev
```
which is invalid for MacOS. We need to use `brew` packaging instead (`brew install libjpeg`). So we need to show last line of the message based on OS.

kgorecki:
Fixes #604

### DIFF
--- a/examples/file_operation/makefile
+++ b/examples/file_operation/makefile
@@ -19,6 +19,20 @@ all: example_file_operation
 clean:
 	$(RM) example_file_operation
 
+ifeq ($(OS),Windows_NT) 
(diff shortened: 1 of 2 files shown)


ihhub:
@kgorecki , thank you so much for the fix!

PR #690, opened 2022-10-26
Fixed png and jpeg libraries detection for macos

### LINKED ISSUE #605: Png and jpeg detection in MacOS makefile is incorrect
In makefiles we have these lines of code:
```
PNG_NOT_EXISTS = $(shell g++ -lpng 2>&1 > /dev/null | grep 'cannot find' | wc -c)
JPEG_NOT_EXISTS = $(shell g++ -ljpeg 2>&1 > /dev/null | grep 'cannot find' | wc -c)
```
These code doesn't work on MacOS with installed **libpng** and **libjpeg** libraries. We need to fix it.

kgorecki:
Fixes #605

### DIFF
--- a/examples/file_operation/makefile
+++ b/examples/file_operation/makefile
@@ -1,8 +1,6 @@
 ##
 # Flags for C++ compiler
 ##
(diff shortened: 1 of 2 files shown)


ihhub:
@kgorecki , thank you so much for this fix!

PR #693, opened 2023-10-31
Adding a cmake config to help find image resources 

### LINKED ISSUE #694: Unable to debug example projects as image paths are in a different location
I love this project, thanks for the work.
I had an issue today on windows where upon getting the source code and running the examples, paths to the images in data were not correct. Potentially I'm just missing something in the build instructions, I have a PR with one potential fix with CMake but wanted to ask about it first. [https://github.com/ihhub/penguinV/pull/693](url)

matthewroberthenderson:
When debugging an an example app after first configuring the project the image paths cannot be located. Adding a simple in file could also be used for other relative project paths.

close #694


ihhub:
Hi @matthewroberthenderson , the provided by you solution fails on Linux (see [this](https://github.com/ihhub/penguinV/actions/runs/6705748674/job/18254385428?pr=693)).

Let's understand the root of the problem: I assume that images are not being copied together with source files during CMake configuration. Therefore, I suggest to add file copying part in CMake files and use relative path for images without modifying source files.

matthewroberthenderson:
@ihhub Yep that makes more sense to me thanks for this. I'll update my PR.

ihhub (approved):

ihhub:
@matthewroberthenderson , thank you very much for your contribution to the project!

matthewroberthenderson:
> @matthewroberthenderson , thank you very much for your contribution to the project!

Very happy to. Really love this project, thanks for putting it out there.

